PR #1, opened 2021-06-16
Add nav listener unregister on app unmount

gkarat:
Applicate the identical patch that was made for `frontend-starter-app`: https://github.com/RedHatInsights/frontend-starter-app/pull/513


AllenBW:
woops, already slammed in a commit before seeing this 😢 😬 🙈

PR #5, opened 2021-07-01
Draft: Configure Redux store, link app with Smart Proxy API

gkarat:
Fixes https://issues.redhat.com/browse/CCXDEV-5298

Mostly copy-paste from the RHEL Advisor with minor differences. 

- [ ] Setup first version of redux store (initialState, reducers, actions, constants).
- [ ] App is authorised to perform requests to the Smart Proxy API (for now, let's utilize production environment everywhere, cloud.redhat.com/api), e.g. [GET /clusters/{clusterId}/report](https://cloud.redhat.com/docs/insights-results-aggregator/insights-results-aggregator#operations-prod-getReportForCluster).


package.json
```diff
@@ -28,14 +28,17 @@
     "@redhat-cloud-services/frontend-components": "3.2.6",
     "@redhat-cloud-services/frontend-components-notifications": "3.2.2",
     "@redhat-cloud-services/frontend-components-utilities": "3.2.2",
+    "axios": "^0.21.1",
```
AllenBW, line 31: locking down the dep version here is recommended

AllenBW (changes requested):
just the one thing, n update the lockfile 🤗

package.json
gkarat: @AllenBW, what's the reason for locking it? Just out of curiosity

AllenBW: so we know the exact version to expect everywhere

AllenBW (approved):
⚡

PR #8, opened 2021-07-05
Add cluster details page (1/2)

gkarat:
Fixes https://issues.redhat.com/browse/CCXDEV-4984

Forked from https://github.com/RedHatInsights/ocp-advisor-frontend/pull/6

PR #9, opened 2021-07-09
Add cluster details page (2/2)

gkarat:
Fixes https://issues.redhat.com/browse/CCXDEV-4984

Based on top of https://github.com/RedHatInsights/ocp-advisor-frontend/pull/8. Please, merge this PR only after https://github.com/RedHatInsights/ocp-advisor-frontend/pull/8.

Let's continue on the Cluster Details view and finalize related components.

- [x] Make Cluster Details view work with the cluster-ID provided 🚀🧑🏼‍🚀
- [x] Add translation scripts (formatjs) 🌎

Screenshot from 2021-07-19 16-44-18


src/PresentationalComponents/ClusterDetailHead/ClusterDetailHead.js
```diff
@@ -0,0 +1,43 @@
...
+  <React.Fragment>
+    <Grid md={6}>
+      <GridItem>
+        <div>
+          <span>UUID</span>
+          <span>
+            {clusterFetchStatus !== 'fulfilled' ? (
+              <Skeleton size="md" />
+            ) : (
+              cluster.cluster_id && 'Unknown'
```
AllenBW, line 17: Can  be a translatable string here

src/PresentationalComponents/ClusterDetailHead/ClusterDetailHead.js
```diff
@@ -0,0 +1,43 @@
...
+import PropTypes from 'prop-types';
+
+import { Grid, GridItem } from '@patternfly/react-core';
+import Skeleton from '@redhat-cloud-services/frontend-components/Skeleton';
+
+const ClusterDetailHead = ({ cluster, clusterFetchStatus }) => (
+  <React.Fragment>
+    <Grid md={6}>
+      <GridItem>
+        <div>
```
AllenBW, line 11: <GridItem> should be sufficient container... shouldn't need div? 🤔

src/PresentationalComponents/ClusterDetailHead/ClusterDetailHead.js
AllenBW: ooo and everywhere else, last seen etc, not gonna comment on each of em, just the one

src/PresentationalComponents/Clusters/ClusterDetails.js
```diff
@@ -0,0 +1,68 @@
...
+  match: PropTypes.any,
+};
+
+const mapStateToProps = ({ cluster, clusterFetchStatus, props }) => ({
+  cluster,
+  clusterFetchStatus,
+  ...props,
+});
+
+export default routerParams(connect(mapStateToProps, null)(ClusterDetails));
```
AllenBW, line 68: Rather than using connect consider using `useSelector` (for store props) `usedispatch` (for dispatching actions) here's n example https://github.com/RedHatInsights/insights-advisor-frontend/blob/prod-stable/src/PresentationalComponents/SystemsTable/SystemsTable.js#L44

src/PresentationalComponents/Loading/Loading.js
```diff
@@ -0,0 +1,15 @@
+import { Card } from '@patternfly/react-core/dist/js/components/Card/Card';
+import { CardBody } from '@patternfly/react-core/dist/js/components/Card/CardBody';
+import { List } from 'react-content-loader';
+import React from 'react';
+const Loading = () => (
+  <>
```
AllenBW, line 6: Either use <React.Fragment> or <> but not both, consistency is keeeeeyyyyy tbh tho, here, i dont think yah need it, card is sufficient

src/PresentationalComponents/ReportDetails/ReportDetails.js
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
```
AllenBW, line 1: ok this whole thing is a federated module that advisor advertises, yah shouldn't keep it here, otherwise that whole drifting of design will happen real fast, lets talk about how yah can consume this from advisor (or checkout https://github.com/RedHatInsights/content-preview which consumes it already or the inventory app)

AllenBW (changes requested):
Real quick, gotta update the test routine in package.json https://app.travis-ci.com/github/RedHatInsights/ocp-advisor-frontend/jobs/523152160#L233****

also gotta run npm run translations... oooo also add that task, check advisor for howwwwwwww (builds translation files)

gkarat:
@AllenBW, thanks for the comments; I will check them out soon. Meanwhile, I've given a try to `@reduxjx/toolkit` and use it to replace the Redux logic with [the last commit](https://github.com/RedHatInsights/ocp-advisor-frontend/pull/9/commits/d3d806010a9d4b87f5b1af045ffc3ffd4abff6f5). It is an interesting package that is considered a better approach than seamless-immutable/immer (actually, immer is utilized by the toolkit).

But I am worried about one thing: what are the consequences of [this change](https://github.com/RedHatInsights/ocp-advisor-frontend/pull/9/commits/d3d806010a9d4b87f5b1af045ffc3ffd4abff6f5#diff-53e22846ad33b36d542d6bf678691a99fbf46fa2f3fea9ddc4175e46c04dd605L39)? I am not really familiar with the r.c.s. `Registry`. Does it mean that the OCP Advisor's store won't be shared across other apps?

gkarat:
... (to my last comment above), just to mention, I haven't yet tested Cluster Details view and redux behavior 😸  -- waiting for a config review from Martin Marosi

AllenBW:
@gkarat totally down with ah different redux flavor! whatever yah wanta do 👯

gkarat:
@AllenBW, I have successfully imported AdvisorReportDetails, but we've got two issues here:

1. We have slightly different fields, so I needed to brute-force and changed 'em when sending properties. I need to address this issue together with API devs.
2. The styling seems to be broken (see the example screenshot)
image. Do you have an idea why it happens? It looks like an SCSS import issue somewhere in between...

gkarat:
Aaand yes, I have copy-pasted and adopted your already existing Advisor table 😄 In the future iterations, we can identify the common stuff and create a federated module - table base for both Advisors. WDYT?

AllenBW:
> ve copy-pasted and adopted your already existing Advisor table 😄 I have left only two filters: description and total risk for now. There is a mess with out categories/tags, so I would leave it for now as a first version.
> 
> In the future iterations, we can identify the common stuff and create a federated module - table base for both Advisors. WDYT?

yeah! we can possibly look at breaking out that table as a fed mod in da future, getting it working first n fast is highest priority it seems... some of the mess yer running into might be a reason not to tho

AllenBW:
> @AllenBW, I have successfully imported AdvisorReportDetails, but we've got two issues here:
> 
>     1. We have slightly different fields, so I needed to brute-force and changed 'em when sending properties. I need to address this issue together with API devs.
> 
>     2. The styling seems to be broken (see the example screenshot)
>        image. Do you have an idea why it happens? It looks like an SCSS import issue somewhere in between...

Hmm this isn't an issue in other places consuming the fed mod, maybe check your config, compare it to others, focus on the styling ingestion differences? 

do yah see any of the style classes present? if yah see em, then thats a different fix if yah dont

as for fields undesired, whats the list of what yah dont need? maybe we can fiddle with it n make em all visible by default, but all able to be hidden?

gkarat:
@AllenBW, PTAL 🙏🏼⚔️🔨

src/Components/Loading/Loading.js
```diff
@@ -3,13 +3,13 @@ import { CardBody } from '@patternfly/react-core/dist/js/components/Card/CardBod
 import { List } from 'react-content-loader';
 import React from 'react';
 const Loading = () => (
-  <>
+  <React.Fragment>
```
AllenBW, line 6: looks like yah can kill this, doesn't do anything, card is a fine root 🤷

src/Components/ReportDetails/ReportDetails.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import AsyncComponent from '@redhat-cloud-services/frontend-components/AsyncComponent';
+
+// Rule fields processing: https://docs.google.com/document/d/1D0ZLD6-2DZqBX1cUHJY2vL2QAD0dIAzY3O3xFNlEsVg/edit?usp=sharing
+
+const ReportDetails = ({ report }) => {
+  console.log({
```
AllenBW, line 9: probably want to kill this console.log 🤔

AllenBW (approved):
Just the two minor observations, that aside see nothing too offensive here! AKA LOOKS FLIPPIN AWESOME @gkarat  wellllll doneeeeeeeee

gkarat:
OK, if no one objects, I am merging it 🐙

PR #17, opened 2021-07-19
Better 404 and empty state messages

gkarat:
Just a minor work on edge cases and their empty states.


gkarat:
Screenshot from 2021-07-20 11-22-37
Screenshot from 2021-07-20 11-22-24
Screenshot from 2021-07-20 11-22-07

gkarat:
@AllenBW, I will merge it now, as it is a minor change. Still, if there is something suspicious, feel free to comment 😸

PR #18, opened 2021-07-20
Use RTK queries for API requests

gkarat:


gkarat:
Tagging you as a reviewer to share my experience with RTK queries. Tbh, it looks really sweet and eliminates a lot of code 👾

gkarat:
And by this patch we basically have the app fully integrated into redux/toolkit 🥳

src/Components/ClusterRules/ClusterRules.js
```diff
@@ -385,62 +384,43 @@ const ClusterRulesTable = ({ cluster }) => {
...
+            isStickyHeader
+          >
+            <TableHeader />
+            <TableBody />
+          </Table>
+          {results === 0 && (
+            <Card>
+              <CardBody>
+                <MessageState
+                  title="No matching recommendations found"
```
AllenBW, line 408: probably gonna want to use intl for this string

AllenBW:
> Tagging you as a reviewer to share my experience with RTK queries. Tbh, it looks really sweet and eliminates a lot of code 👾

Oh this is heckin neat! excited to see how it plays out... 

i LOVE how all states are destructed from `useGetClusterByIdQuery(match.params.clusterId);` also the builtin refresh... does it take params that modify the previous call?

gkarat:
> > Tagging you as a reviewer to share my experience with RTK queries. Tbh, it looks really sweet and eliminates a lot of code space_invader
> 
> Oh this is heckin neat! excited to see how it plays out...
> 
> i LOVE how all states are destructed from `useGetClusterByIdQuery(match.params.clusterId);` also the builtin refresh... does it take params that modify the previous call?
> 
> OOOOOO ALSO THE `isUninitialized` PARAM now thats real tempting

hmm, it seems to be not supported. Docs aren't saying anything about this func, and in VSCode it's marked as `() => void()`. Also, I tried to intentionally call `refetch` again with the random string parameter, but the new request uses the initial ID given in `useGetClusterByIdQuery`.

AllenBW (approved):
LOOKS LIKE FUNNNN maybe i'll give it a whirl when implementing pathways 🤔

gkarat:
> LOOKS LIKE FUNNNN maybe i'll give it a whirl when implementing pathways

yeah, give it a try! I think this thing will get popular really quickly. I am only concerned about result bundle sizes... will once compare the before/after approaches in my free time

PR #20, opened 2021-07-22
Show cluster display name

gkarat:
Clusters can have custom names set up by administrators. AMS API is the place where Advisor can obtain this data.


gkarat:
example 🧑‍🔧 
Screenshot from 2021-07-22 17-23-13

PR #21, opened 2021-07-27
Create basic unit tests

gkarat:
Fixes https://issues.redhat.com/browse/CCXDEV-5407

- [x] Decouple components to comply with SRP
- [x] For new (semi-) pure components, create basic unit tests 🛸

### DIFF
--- a/config/setupTests.js
+++ b/config/setupTests.js
@@ -1,6 +1,6 @@
+import React from 'react';
 import { configure, mount, render, shallow } from 'enzyme';
 import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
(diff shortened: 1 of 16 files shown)


gkarat:
@AllenBW, could you have a quick look at it? 🐱

gkarat:
will merge it now as I need to proceed and don't want to create git conflicts; if something looks suspicious, we can tackle it in a separate PR 🥷

btw, the enzyme library looks a bit hard to use with the functional components wrapped by contexts / using context hooks inside. I am looking towards https://testing-library.com/docs/react-testing-library/intro/ in the future instead

PR #26, opened 2021-08-20
Add recommendation details page (without affected clusters table)

gkarat:
Fixes https://issues.redhat.com/browse/CCXDEV-5495
Mock-up https://marvelapp.com/prototype/g70a1he/screen/79082637

Screenshot from 2021-08-24 16-58-45
Screenshot from 2021-08-24 16-59-03


gkarat:
Waiting for https://github.com/RedHatInsights/insights-advisor-frontend/pull/861 before I can properly test this PR.

gkarat:
Summoning @AllenBW ⛩️🌟 PTAL

gkarat:
Right now, we cannot make the "affected clusters" component work: we don't have a [rule:affected clusters] mapping is not yet implemented in the backend. https://issues.redhat.com/browse/CCXDEV-5088 should solve this.

gkarat:
@doruskova, I need your review too ideally 😸

gkarat:
The risk of change value is also not available now; that's why the field is hidden. Also, the feedback form is not shown either, as right now, we have feedback for the cluster + rule set (the feature will probably appear on the cluster details page).

doruskova:
@gkarat I'm sorry that I didn't give you an empty state for it. In RHEL they're using an empty state from Inventory and it doesn't make sense for our use case. So, I created this one for the successful state: 
03 - Details page - empty-state (no clusters)



And this is the error state: 
03 - Details page - error when loading recommendation (empty state)

Fewwy (commented):
LGTM

doruskova:
> LGTM

Nope! The visual is not correct.

gkarat:
@doruskova, thank you, Marie! Any chance we could adopt [this text from my screenshot](https://user-images.githubusercontent.com/31385370/130518131-87892458-9dc5-46cf-8178-5d2703eb6e13.png)? The reason is https://github.com/RedHatInsights/ocp-advisor-frontend/pull/26#issuecomment-904399871. Or we rather [don't show the whole page and show error ](https://marvelapp.com/prototype/g70a1he/screen/81633703)until the affected clusters feature is effective?

doruskova:
@gkarat Unfortunately no, it breaks the consistency. Do I understand correctly that it's going to affect only the affected clusters part? Why don't you show the upper part in that case? Sorry for my questions, haha :D

Fewwy:
> > LGTM
> 
> Nope! The visual is not correct.

I was looking at the code :D Missed the visual difference

gkarat:
> @gkarat Unfortunately no, it breaks the consistency. Do I understand correctly that it's going to affect only the affected clusters part? Why don't you show the upper part in that case? Sorry for my questions, haha :D
> 
> Ooh, you're talking about the first mock, right?

yeah, I was talking the first mock: whether we could show the upper part, but always have the same message for the affected clusters table in a sense "sorry, we couldn't obtain info about the affected clusters."

doruskova:
Ok, so we decided to use platform error state when the url is wrong. 
Snímek obrazovky 2021-08-24 v 10 28 09

Updated title for no clusters
03 - Details page - empty-state (no clusters)

src/Components/Breadcrumbs/Breadcrumbs.js
```diff
@@ -46,5 +46,6 @@ export const Breadcrumbs = ({ current, match }) => {
 
 Breadcrumbs.propTypes = {
   current: PropTypes.string,
-  match: PropTypes.object,
+  match: PropTypes.object.isRequired,
+  intl: PropTypes.object.isRequired,
```
AllenBW, line 50: Dont need this proptype

src/Components/Recommendation/index.js
```diff
@@ -0,0 +1,23 @@
+import React, { useEffect } from 'react';
+import { useIntl } from 'react-intl';
+
+import routerParams from '@redhat-cloud-services/frontend-components-utilities/RouterParams';
```
AllenBW, line 4: prefer the use of react-router-dom to routerParams

src/Components/Recommendation/index.js
```diff
@@ -0,0 +1,23 @@
...
+
+import { Recommendation } from './Recommendation';
+import { useGetRuleByIdQuery } from '../../Services/SmartProxy';
+import { getPluginName } from '../../Utilities/RuleName';
+
+const RecommendationWrapper = routerParams(({ match }) => {
+  const intl = useIntl();
```
AllenBW, line 11: Same thing as before, prefer invocation of intl where its being used rather than passing the basket around

src/Messages.js
```diff
@@ -216,4 +211,54 @@ export default defineMessages({
     defaultMessage:
       'Indicates this recommendation will not be shown for the cluster.',
   },
+  rulesDetailsPubishdate: {
```
AllenBW, line 214: be sure to run yer translation script when adding stuff to messages.js

AllenBW (commented):
Some thoughts! lookin mighty strong, WOOH FEDERATED MODULE FOR THE WIN

src/Components/Breadcrumbs/Breadcrumbs.js
```diff
@@ -46,5 +46,6 @@ export const Breadcrumbs = ({ current, match }) => {
 
 Breadcrumbs.propTypes = {
   current: PropTypes.string,
-  match: PropTypes.object,
+  match: PropTypes.object.isRequired,
```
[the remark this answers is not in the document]
gkarat: well, my point here was to keep components pure (in the terms of being independent of contexts and receiving data only from props). It helps to write unit tests faster (you don't need to set up a context for the unit test, you just send what it needs via props). So you can see from time to time that I make wrappers for some components that consume all the contexts and then pass it to the inner component via props.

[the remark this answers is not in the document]
gkarat: well, however, I still have made an exception for `intl`, see `config/setupTests.js:13` 😄

src/Components/Recommendation/index.js
gkarat: @AllenBW, why the @r-c-s's version is not preferred by you?

r-c-s: Wrong r-c-s? I have nothing to do with this project.

r-c-s: @gkarat Wrong r-c-s? I have nothing to do with this project.

gkarat: ooops, sorry, wrong mention 😅

src/Components/Breadcrumbs/Breadcrumbs.js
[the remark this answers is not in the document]
AllenBW: Sure! I get that, not my personal preference, but s all good, thanks for clarifying 🥰

src/Components/Recommendation/index.js
AllenBW: Another personal preference, in this case ah belief that the package will be kinder to the component lifecycle, it's also mo powerful, (fully featured)

gkarat:
OK, let's try it out in QA/beta...

PR #28, opened 2021-08-25
Fix eslint errors failing master build

gkarat:


gkarat:
just a minor fix to heal master's build. No functional changes

PR #51, opened 2021-11-03
Wrap app with auth check

gkarat:
Fixes https://issues.redhat.com/browse/CCXDEV-6237.

Introduces authentication check and wrapper for Advisor. Once the user has authenticated within insights chroming, the content of Advisor is exposed. The redux store is also now not initialized, while the user has not passed auth check.

Screenshot 2021-11-03 at 13-51-57 0f9b7635-b5bf-4a6f-be8e-644cfb60a794 - Clusters - OCP Advisor Red Hat Insights

### DIFF
--- a/compiled-lang/en.json
+++ b/compiled-lang/en.json
@@ -48,6 +48,8 @@
   "none": "None",
   "oneOrMore": "1 or more",
   "performance": "Performance",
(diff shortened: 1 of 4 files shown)


gkarat:
@Fewwy, please, take a look 🔍

PR #54, opened 2021-11-05
Remove select feature; utilize stage AMS API for pre-prod envs

gkarat:
- For the single cluster page, remove the bulk select and all select-related features from the recommendations table. (See screenshot 1)
- When deployed in pre-prod envs (e.g. qa/beta), use AMS API in the stage environment.
- Change empty state body message for the rules table (screenshot 2).

Screenshot 2021-11-05 at 11-50-39 576968e7-b50e-4723-8952-81b0477b6f4f - Clusters - OCP Advisor Red Hat Insights

Screenshot 2021-11-05 at 12-03-37 OCP Advisor Insights


doruskova:
@gkarat Thank you! Looks good.

PR #56, opened 2021-11-08
Add sorting to columns in the recommendations list

Fewwy:
Functionality:
1)Added sorting to the columns on the Recommendation List page
2)Added parameters to the Recommendation list object it the state in the Filters slice of redux store
3)Wrote "onSort" function and connected it to the Recommenadtion List object

Tests:
1)Wrote cypress test that are checking if the table columns are rendered and sortable
2)Wrote cypress tests that are checking if the sorting works correctly


src/Components/RecsListTable/index.js
```diff
@@ -296,6 +299,35 @@ const RecsListTable = () => {
     },
   ];
 
+  const onSort = (event, index, direction) => {
+    const sortedRecommendations = {
+      2: 'name',
```
gkarat, line 304: Why does the indexing start with 2?

src/Components/RecsListTable/index.js
```diff
@@ -53,6 +54,8 @@ const RecsListTable = () => {
   const page = filters.offset / filters.limit + 1;
   const [filteredRows, setFilteredRows] = useState([]);
   const [displayedRows, setDisplayedRows] = useState([]);
+  const [recommendations, setRecommendations] = useState(recs);
```
gkarat, line 57: Do you really need to keep another array of recs? Why you can't modify `filteredRows` instead?

src/Components/RecsListTable/index.js
```diff
@@ -296,6 +299,35 @@ const RecsListTable = () => {
     },
   ];
 
+  const onSort = (event, index, direction) => {
+    const sortedRecommendations = {
+      2: 'name',
+      3: 'added',
+      4: 'total risk',
```
gkarat, line 306: Is it a correct parameter name? It looks like it has a missing underscore.

src/Components/RecsListTable/index.js
```diff
@@ -53,6 +54,8 @@ const RecsListTable = () => {
   const page = filters.offset / filters.limit + 1;
   const [filteredRows, setFilteredRows] = useState([]);
   const [displayedRows, setDisplayedRows] = useState([]);
+  const [recommendations, setRecommendations] = useState(recs);
+  const [sortBy, setSortBy] = useState({});
```
gkarat, line 58: Please, keep all filtering/sorting parameters in the global redux store. In the end, you should access sorting parameters in the same way as `filters` with `useSelector`.

src/Components/RecsListTable/index.js
Fewwy: Good catch

src/Components/RecsListTable/index.js
Fewwy: Because I don't understand how it work, how should I change them?

src/Components/RecsListTable/index.js
Fewwy: I will do that as a part of the refactor as soon as the sorting works without bugs

src/Components/RecsListTable/index.js
```diff
@@ -62,107 +64,118 @@ const RecsListTable = () => {
...
-                      {value?.description || value?.rule_id}{' '}
-                    </Link>
-                    <RuleLabels rule={value} />
-                  </span>
-                ),
-              },
-              {
-                title: value?.publish_date ? (
-                  <DateFormat
+  const buildFilteredRows = (allRows, filters, index) => {
```
gkarat, line 67: Is it expected that you keep some of the `buildFilteredRows` calls without the third parameter? I mean, here https://github.com/RedHatInsights/ocp-advisor-frontend/blob/9ead83983159bf292f726e51564bac6248393f2b/src/Components/RecsListTable/index.js#L63, the function does not provide `index`. What would happen when `index` is undefined?

gkarat: I would also provide a sorting direction parameter (or just a single `sortBy` object) and keep direction logic inside the function. Right now, you decide whether to reverse the rows here https://github.com/RedHatInsights/ocp-advisor-frontend/blob/9ead83983159bf292f726e51564bac6248393f2b/src/Components/RecsListTable/index.js#L313

src/Components/RecsListTable/index.js
```diff
@@ -296,6 +309,18 @@ const RecsListTable = () => {
     },
   ];
 
+  const onSort = (_e, index, direction) => {
+    const sortRecsDirection =
+      direction === SortByDirection.asc
+        ? buildFilteredRows(recs, filters, index)
```
gkarat, line 315: Again, what about calling just buildDisplayedRows and providing the sorting parameters to this function? So that the function 1) sorts the rows that were already pre-filtered by buildFilteredRows, and 2) makes a slice of this array to meet pagination params (already implemented).

src/Components/RecsListTable/index.js
```diff
@@ -296,6 +309,18 @@ const RecsListTable = () => {
...
+        : buildFilteredRows(recs, filters, index).reverse();
+    setSortBy({
```
gkarat, line 317: Just to add: once you move the sorting params to the redux store, you should call dispatch instead of the set state function.

gkarat (changes requested):

src/Services/Filters.js
```diff
@@ -29,10 +31,20 @@ const filters = createSlice({
     updateRecsListFilters(state, action) {
       state.recsListState = action.payload;
     },
+    sortTableIndex(state, action) {
```
gkarat, line 34: Setters or "update actions" should have an appropriate name in the style "[**update**|**set**]SortIndex" or similar.

src/Components/RecsListTable/index.js
```diff
@@ -63,117 +72,129 @@ const RecsListTable = () => {
...
+          : 0
+      )
       .slice(
         filters.limit * (page - 1),
         filters.limit * (page - 1) + filters.limit
       )
       .flatMap((row, index) => {
         const updatedRow = [...row];
         row[1].parent = index * 2;
+        direction === SortByDirection.asc ? updatedRow : updatedRow.reverse();
```
gkarat, line 197: Why you would ever change the `direction` **parameter** of the `buildDisplayedRows` function?..

What is the point of reversing one single row? Modifying flatMap here is most probably not required at all.

src/Components/RecsListTable/index.js
```diff
@@ -52,9 +55,15 @@ const RecsListTable = () => {
   const page = filters.offset / filters.limit + 1;
   const [filteredRows, setFilteredRows] = useState([]);
   const [displayedRows, setDisplayedRows] = useState([]);
+  const setRecListSortIndex = (filters) =>
```
gkarat, line 58: I would rather not wrap the dispatch method with another function at all. Just call dispatch where the update of these parameters is needed.

src/Components/RecsListTable/index.js
gkarat: From what I can see, I would call reverse function right after `.sort` https://github.com/RedHatInsights/ocp-advisor-frontend/pull/56/files#diff-4a5517c05a9afe6cc744c7b111538de5db09e9c5fd1a2aac0d315dd51a1f1da4R181

gkarat (changes requested):

src/Services/Filters.js
```diff
@@ -12,6 +12,8 @@ export const RECS_LIST_INITIAL_STATE = {
   limit: 20,
   offset: 0,
   impacting: ['true'],
+  sortIndex: 0,
```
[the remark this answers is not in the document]
Fewwy: Should I use -1 instead? Or just null?

Fewwy:
@gkarat PTAL

src/Components/RecsListTable/index.js
```diff
@@ -296,6 +333,11 @@ const RecsListTable = () => {
     },
   ];
 
+  const onSort = (_e, index, direction) => {
+    dispatch(SortRecsTableIndex(index));
```
gkarat, line 337: A nitpick: as you don't change these index and direction values separately anywhere in the code, it could be better to keep only one action called `updatePagination` or update the pagination values using existing `updateFilters` like

```suggestion
    dispatch(
      updateFilters({ ...filters, sortIndex: index, sortDirection: direction })
    );
```

src/Components/RecsListTable/index.js
```diff
@@ -34,6 +35,8 @@ import messages from '../../Messages';
 import {
   RECS_LIST_INITIAL_STATE,
   updateRecsListFilters as updateFilters,
+  updateRecsListSortIndex as SortRecsTableIndex,
```
gkarat, line 38: A nitpick: `SortRecsTableIndex` is a bid odd choice for the function name. I quite recommend to be consistent within the code and use 1) snake case, 2) verbs "update", "set" for such functions. E.g., consider
```suggestion
  updateRecsListSortIndex as updateSortIndex,
```

src/Components/RecsListTable/index.js
```diff
@@ -54,119 +57,153 @@ const RecsListTable = () => {
...
+          ? 1
+          : secondItem[0].rule[sortedRecommendations[index - 1]] >
+            firstItem[0].rule[sortedRecommendations[index - 1]]
+          ? -1
+          : 0;
+      });
+    } else {
+      sortingRows
+        .sort((firstItem, secondItem) =>
+          firstItem[0].rule[sortedRecommendations[index - 1]] >
```
gkarat, line 196: A nitpick: two identical comparison within the if-else block: follow DRY

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -98,4 +98,128 @@ describe('recommendations list table', () => {
...
+
+  it('Recommendation table sort the data by Name second sort', () => {
+    cy.get(RECS_LIST_TABLE)
+      .get('span[class=pf-c-table__sort-indicator]')
+      .first()
+      .click({ force: true })
+      .click({ force: true });
+    cy.get('td[data-label=Name]')
+      .find('a')
+      .should(($a) => {
```
gkarat, line 139: The same story: please see my comment above and update for the rest of the tests if applicable.

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -98,4 +98,128 @@ describe('recommendations list table', () => {
...
+        expect($span[0]).to.have.text('Moderate');
+      });
+  });
+
+  it('Recommendation table sort the data by Total Risk second sort', () => {
+    cy.get(RECS_LIST_TABLE)
+      .get('span[class=pf-c-table__sort-indicator]')
+      .eq(2)
+      .click({ force: true })
+      .click({ force: true });
```
gkarat, line 193: A nitpick: what about https://docs.cypress.io/api/commands/dblclick? 😄

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -98,4 +98,128 @@ describe('recommendations list table', () => {
...
+      .eq(3)
+      .click({ force: true });
+    cy.get('td[data-label=Clusters]')
+      .find('div')
+      .should(($div) => {
+        expect($div[0]).to.have.text('1');
+      });
+  });
+
+  it('Recommendation table sort the data by Clusters second sort', () => {
```
gkarat, line 213: I propose merging "double-click" tests with the "single-click" tests: it does not make much sense to separate the scenarios IMHO.

gkarat (changes requested):
Hey, generally looks fine, and the feature works, thank you 🥇. Could you take a look at a couple of things before we merge? Ones that are "nitpicks" are just suggestions: we can keep/fix them later; it's up to you to decide.

src/Components/RecsListTable/index.js
Fewwy: I think I tried to do it again and had some issues with the fact that data sent to the state.recsListState.sortDirection was the same index as in the state.recsListState.sortIndex and not "desc"/"asc" strings. 

Let me try to implement that again

gkarat (approved):

gkarat:
LGTM, thank you 🦸🏼

PR #59, opened 2021-11-18
Use cypress-ct for component testing; refactor table components

gkarat:
The patch
- translates some of the snapshot tests into the **component tests**[^1] written using the cypress-ct library,
- for now, it removes the support of the enzyme library since the current testing suite already fulfills the essential unit testing needs (Jest as an assertion library for testing JS functions, cypress component testing for testing isolated components).
- refactors "table components" (e.g., affected clusters, cluster reports no longer have a side effect -- sending a network request) and simplifies existing cypress-ct tests

This is also a strive to give up the snapshot testing in favor of more user-oriented testing while testing components in isolation. Inspired by [^2], [^3].

[^1]: https://css-tricks.com/testing-vue-components-with-cypress/#a-little-groundwork
[^2]: https://kentcdodds.com/blog/testing-implementation-details
[^3]: https://kentcdodds.com/blog/why-i-never-use-shallow-rendering


gkarat:
Found that cypress-ct tests are not included in the tests coverage. Opened https://github.com/RedHatInsights/ocp-advisor-frontend/issues/60.

PR #67, opened 2021-12-08
Add systems (clusters) page

gkarat:
Fixes https://issues.redhat.com/browse/CCXDEV-6312

- [x] Comply with mockups
- [x] Filters working correctly
- [x] Pagination working correctly
- [x] Sorting working correctly
- [x] Show cluster display name instead of ID if available
- [x] Names lead to single cluster page 
- [x] Tests written
- [x] Merge utilities and functions common with other tables (e.g., recommendations list)
- [x] Screenshots provided

Screenshot 2021-12-09 at 16-05-53 OCP Advisor Insights
Screenshot 2021-12-09 at 16-05-39 OCP Advisor Insights
Screenshot 2021-12-09 at 16-05-31 OCP Advisor Insights
Screenshot 2021-12-09 at 16-17-18 OCP Advisor Insights
Screenshot from 2021-12-09 16-06-37

### DIFF
--- a/src/Components/ClustersListTable/ClustersListTable.js
+++ b/src/Components/ClustersListTable/ClustersListTable.js
@@ -24,6 +24,8 @@ import {
 import {
   CLUSTERS_LIST_COLUMNS,
   CLUSTER_FILTER_CATEGORIES,
+  CLUSTER_LAST_CHECKED_CELL,
+  CLUSTER_NAME_CELL,
 } from '../../AppConstants';
 import {
   buildFilterChips,
@@ -34,14 +36,15 @@ import Loading from '../Loading/Loading';
 import messages from '../../Messages';
 import { ErrorState, NoMatchingClusters } from '../MessageState/EmptyStates';
 
-const ClustersListTable = ({ query }) => {
+const ClustersListTable = ({
+  query: { isError, isUninitialized, isFetching, isSuccess, data },
+}) => {
   const intl = useIntl();
   const dispatch = useDispatch();
   const updateFilters = (payload) =>
     dispatch(updateClustersListFilters(payload));
   const filters = useSelector(({ filters }) => filters.clustersListState);
 
-  const { isError, isUninitialized, isFetching, isSuccess, data } = query;
   const clusters = data?.data || [];
   const page = filters.offset / filters.limit + 1;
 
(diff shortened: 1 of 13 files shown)


gkarat:
@doruskova, @RedHatInsights/ui-reviewers, @RedHatInsights/advisor-ui-reviewers, @ryelo, @Fewwy, here is a major patch for the OCP Advisor that adds a new page "Clusters" (with the RHEL's analogy, "Systems" page). If you could at least quickly go through the patch, your input would be quite valuable 🚀 (you can still ping me/take a look at Jira/mock-ups for more info)

src/Components/ClustersListTable/ClustersListTable.js
```diff
@@ -0,0 +1,247 @@
...
+          addFilterParam(CLUSTER_FILTER_CATEGORIES.hits.urlParam, values),
+        value: filters.hits,
+        items: CLUSTER_FILTER_CATEGORIES.hits.values,
+      },
+    },
+  ];
+
+  const activeFiltersConfig = {
+    deleteTitle: intl.formatMessage(messages.resetFilters),
+    // TODO: utilize buildFilterChips in RecsListTable
```
gkarat, line 143: remove

src/Components/Common/Tables.js
```diff
@@ -39,4 +51,133 @@ const passFilters = (rule, filters) =>
...
+            }
+          >
+            <span>{intl.formatMessage(messages.nA)}</span>
+          </Tooltip>
+        )}
+      </span>,
+    ],
+  }));
+
+const capitalize = (string) => string[0].toUpperCase() + string.substring(1);
```
gkarat, line 113: remove the duplicate from other file

doruskova:
@gkarat Looks great! Thank you!

src/Components/Common/Tables.js
```diff
@@ -39,4 +51,133 @@ const passFilters = (rule, filters) =>
...
+
+const pruneFilters = (localFilters, filterCategories) => {
+  const prunedFilters = Object.entries(localFilters);
+  return prunedFilters.length > 0
+    ? prunedFilters.reduce((arr, item) => {
```
karelhala, line 118: No need to check for `length > 0` if the array is empty it will not reduce it. However if the `localFilters` can be undefined you should check for this. The first and second item is used numberous times, it would be better to use `key` and `item` instead.
```suggestion
  return Object.entries(localFilters ?? {}).reduce((arr, [key, item]) => {
```

src/Components/Common/Tables.js
```diff
@@ -39,4 +51,133 @@ const passFilters = (rule, filters) =>
...
+                    category: 'Name',
+                    chips: [{ name: item[1], value: item[1] }],
+                    urlParam: item[0],
+                  },
+                ]
+              : []),
+          ];
+        } else {
+          return arr;
+        }
```
karelhala, line 166: No need for extra `else` block.

karelhala (commented):
Looking good! Just a few enhancements and we are golden!

src/Components/ClustersListTable/ClustersListTable.js
```diff
@@ -0,0 +1,247 @@
...
+} from '../Common/Tables';
+import Loading from '../Loading/Loading';
+import messages from '../../Messages';
+import { ErrorState, NoMatchingClusters } from '../MessageState/EmptyStates';
+
+const ClustersListTable = ({ query }) => {
+  const intl = useIntl();
+  const dispatch = useDispatch();
+  const updateFilters = (payload) =>
+    dispatch(updateClustersListFilters(payload));
```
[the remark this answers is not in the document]
gkarat: @karelhala, do you really think that simple one-line functions like the `dispatch` here really require a `useCallback` wrapping? Furthermore, the function `updateFilters` is not listed in any of the `useEffect` dependencies == the re-calculating of the instance of this function does not trigger any further re-renders.

But this is an interesting question, since I have recently read this article https://dmitripavlutin.com/dont-overuse-react-usecallback/. It turns out, the `useCallback` is often overused. Just FYI.

src/Components/ClustersListTable/ClustersListTable.js
```diff
@@ -0,0 +1,247 @@
...
+  }, [data, filters]);
+
+  const buildFilteredRows = (allRows, filters) => {
+    return mapClustersToRows(
+      allRows.filter((cluster) => passFiltersCluster(cluster, filters))
+    );
+  };
+
+  const buildDisplayedRows = (rows, index, direction) => {
+    const sorted = [...rows];
```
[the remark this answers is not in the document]
gkarat: @karelhala, you are true, but I also do the in-place sort logic within the `buildDIsplayedRows` function (see the next lines).

src/Components/ClustersListTable/ClustersListTable.js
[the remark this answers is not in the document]
Hyperkid123: @gkarat it depends on the complexity of the function you want to memorize. In this case, I agree, the `updateClustersListFilters` function is very simple. Adding the hook around it probably just makes the code slower.

PR #68, opened 2021-12-12
Added expand/collapse all button with icon to the Primary toolbar

Fewwy:
Described status expanded true/false in the state using useState
used property expandAll of PrimaryToolbar to pass it function collapseAll
image

### DIFF
--- a/src/Components/RecsListTable/RecsListTable.js
+++ b/src/Components/RecsListTable/RecsListTable.js
@@ -83,7 +83,7 @@ const RecsListTable = ({ query }) => {
       .filter((rule) => passFilters(rule, filters))
       .map((value, key) => [
         {
-          isOpen: false,
+          isOpen: isAllExpanded ? true : false,
           rule: value,
           cells: [
             {
--- a/src/Components/RecsListTable/RecsListTable.js
+++ b/src/Components/RecsListTable/RecsListTable.js
@@ -3,7 +3,9 @@ import { useDispatch, useSelector } from 'react-redux';
 import { useIntl } from 'react-intl';
 import { Link } from 'react-router-dom';
 import PropTypes from 'prop-types';
-
+import { Button } from '@patternfly/react-core/dist/js/components/Button';
+import AngleRightIcon from '@patternfly/react-icons/dist/js/icons/angle-right-icon';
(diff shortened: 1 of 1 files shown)


src/Components/RecsListTable/RecsListTable.js
```diff
@@ -497,6 +512,7 @@ const RecsListTable = ({ query }) => {
         />
       )}
       <PrimaryToolbar
+        expandAll={{ isAllExpanded, onClick: collapseAll }}
```
gkarat, line 515: That is great that you used this simple property of the toolbar. However, I have spotted one bug. 

The `isAllExpanded` state parameter is not updated once you change the any of the filters/pagination. Imagine, that you have expanded all the rules, and if you navigate to the next page or apply any of the filters, you will see that the arrow button is still effective (= pointing downwards), but all the rows are hovered back.

I think, the only thing you need to enhance this `useEffect`: https://github.com/RedHatInsights/ocp-advisor-frontend/blob/c8fa96f39761ff0a05a93bfa0e3783a5a1ed2445/src/Components/RecsListTable/RecsListTable.js#L64

gkarat (changes requested):

gkarat:
Overall LGTM, but it seems we better treat this use case I have mentioned and then we can merge it.

Fewwy:
@gkarat PTAL Should I add something else or is that's enough?

src/Components/RecsListTable/RecsListTable.js
Fewwy: Used terner operator in buildFilteredRows to have the same effect as you want
isOpen: isAllExpanded ? true : false

gkarat (approved):
Looks great, thank you, mergin'

src/Components/RecsListTable/RecsListTable.js
```diff
@@ -83,7 +83,7 @@ const RecsListTable = ({ query }) => {
       .filter((rule) => passFilters(rule, filters))
       .map((value, key) => [
         {
-          isOpen: false,
+          isOpen: isAllExpanded ? true : false,
```
gkarat, line 86:
```suggestion
          isOpen: isAllExpanded,
```

gkarat (approved):

PR #74, opened 2021-12-21
When authenticating user, render spinner instead of the lock icon

gkarat:


gkarat:
@Fewwy, it's very minor one, but still pinging you here for visibility

Fewwy:
@gkarat Looks good to me! :)

PR #76, opened 2021-12-23
Hide disabled rules for single cluster page

Fewwy:
Rewrote API query so we'll request only enabled rules for single cluster page

Removed unnecessary code related to the "status" of the rules on the single cluster page

### DIFF
--- a/src/Components/ClusterRules/ClusterRules.js
+++ b/src/Components/ClusterRules/ClusterRules.js
@@ -185,7 +185,7 @@ const ClusterRules = ({ reports }) => {
               created_at: rule.created_at,
               total_risk: rule.total_risk,
               category: rule.tags,
-              rule_status: rule.disabled ? 'disabled' : 'enabled',
+              rule_status: 'enabled',
             };
             if (key === 'category') {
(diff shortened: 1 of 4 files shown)


src/Components/ClusterRules/ClusterRules.js
```diff
@@ -185,7 +185,7 @@ const ClusterRules = ({ reports }) => {
               created_at: rule.created_at,
               total_risk: rule.total_risk,
               category: rule.tags,
-              rule_status: rule.disabled ? 'disabled' : 'enabled',
+              rule_status: 'enabled',
```
gkarat, line 188: There is no need to keep `rule_status`-related stuff at all. The logic should be removed entirely. Also, take a look at table filters https://github.com/RedHatInsights/ocp-advisor-frontend/blob/939e528298336420a5614434cbbbb63c11f4cf9d/src/Components/ClusterRules/ClusterRules.js#L44
There is a default status filter: 'enabled'.

gkarat (changes requested):

gkarat (approved):

PR #80, opened 2022-01-04
Bump @patternfly/react-* versions; enhance OUIA support

gkarat:
- [x] Upgrade patternfly libraries versions
- [x] Enhance OUIA support across components


ikerreyes:
Add OUIA IDs to enhance testing.

IDs are in lowercase separating words with `-`

PR #82, opened 2022-01-05
Update params that are passed to the RuleDetails federated module

gkarat:
- [x] Fix `impacted_clusters_number` value that was intentionally set to undefined and not propagated to the RuleDetails federated module.
- [x] Pass the `history` object to RuleDetails from the RHEL Advisor so that a custom Router wrapper can be used (relates to https://github.com/RedHatInsights/insights-advisor-frontend/pull/891)

### DIFF
--- a/src/Components/Recommendation/RuleDetails.js
+++ b/src/Components/Recommendation/RuleDetails.js
@@ -1,21 +1,16 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { useIntl } from 'react-intl';
+import { useHistory } from 'react-router-dom';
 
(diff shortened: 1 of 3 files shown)


gkarat:
@RedHatInsights/ui-reviewers, @Fewwy, please take a look, also see the related patch in the RHEL Advisor that could help understand the context.

Hyperkid123 (approved):

gkarat:
I checked with QA deployed module from RHEL Advisor, and it works OK; the message is rendered => I assume the custom router workaround works nicely.

PR #83, opened 2022-01-06
Fixed a bug with the clusters Impacted filter and status All

Fewwy:
There was a bug with filters
After debugging and discussing it with Georgy he advised to remove this line of code, it was a culprit causing this bug.
1. Navigate to the recommendations page
2. Open the filter "Cluster impacted" and leave the "1 or more" selected and then select the "None" (so both are selected)
3. Open the filter status and select "All"

The "1 or more" filter for "Cluster impacted" is removed from the list of filters.

### DIFF
--- a/src/Components/RecsListTable/RecsListTable.spec.ct.js
+++ b/src/Components/RecsListTable/RecsListTable.spec.ct.js
@@ -12,6 +12,8 @@ import { Intl } from '../../Utilities/intlHelper';
 const RECS_LIST_TABLE = 'div[id=recs-list-table]';
 const CHIP = 'div[class=pf-c-chip]';
 const ROW = 'tbody[role=rowgroup]';
+const FILTERS_DROPDOWN = 'ul[class=pf-c-dropdown__menu]';
+const FILTER_TOGGLE = 'span[class=pf-c-select__toggle-arrow]';
 
 // actions
 Cypress.Commands.add('getAllRows', () => cy.get(RECS_LIST_TABLE).find(ROW));
@@ -284,10 +286,6 @@ describe('successful non-empty recommendations list table', () => {
   });
 
   it('the Impacted filters work correctly', () => {
-    const RECS_LIST_TABLE = 'div[id=recs-list-table]';
-    const FILTERS_DROPDOWN = 'ul[class=pf-c-dropdown__menu]';
-    const FILTER_TOGGLE = 'span[class=pf-c-select__toggle-arrow]';
-
     cy.get(RECS_LIST_TABLE).find('button[class=pf-c-dropdown__toggle]').click();
     cy.get(FILTERS_DROPDOWN)
       .contains('Clusters impacted')
@@ -299,8 +297,7 @@ describe('successful non-empty recommendations list table', () => {
       cy.wrap(element);
       element[0].click();
     });
-    cy.get('#pf-random-id-72-false').check({ force: true });
-    cy.get('.pf-c-chip-group__list-item').contains('1 or more');
+    cy.get('.pf-c-check__input').check({ force: true });
 
     cy.get(RECS_LIST_TABLE).find('button[class=pf-c-dropdown__toggle]').click();
     cy.get(FILTERS_DROPDOWN)
--- a/src/Components/RecsListTable/RecsListTable.spec.ct.js
+++ b/src/Components/RecsListTable/RecsListTable.spec.ct.js
@@ -282,6 +282,26 @@ describe('successful non-empty recommendations list table', () => {
       .find('td[data-label=Category]')
       .should('contain', 'Service Availability');
   });
+
+  it('the filters work correctly', () => {
+    const RECS_LIST_TABLE = 'div[id=recs-list-table]';
+    const FILTERS_DROPDOWN = 'ul[class=pf-c-dropdown__menu]';
+    const FILTER_TOGGLE = 'span[class=pf-c-select__toggle-arrow]';
+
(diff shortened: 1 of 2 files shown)


gkarat (changes requested):
Looks good; please, write one test that would cover this scenario. Each bug is a result of testing coverage deficiencies, and it can serve as a good opportunity to improve it.

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -282,6 +282,45 @@ describe('successful non-empty recommendations list table', () => {
       .find('td[data-label=Category]')
       .should('contain', 'Service Availability');
   });
+
+  it('the Impacted filters work correctly', () => {
+    const RECS_LIST_TABLE = 'div[id=recs-list-table]';
```
gkarat, line 287: RECS_LIST_TABLE is already accessible and defined in the outer scope on line 12: https://github.com/RedHatInsights/ocp-advisor-frontend/blob/e8e096dfd4827ac52203391cda8bc899bdeb012b/src/Components/RecsListTable/RecsListTable.spec.ct.js#L12

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -282,6 +282,45 @@ describe('successful non-empty recommendations list table', () => {
...
+    const FILTERS_DROPDOWN = 'ul[class=pf-c-dropdown__menu]';
+    const FILTER_TOGGLE = 'span[class=pf-c-select__toggle-arrow]';
+
+    cy.get(RECS_LIST_TABLE).find('button[class=pf-c-dropdown__toggle]').click();
+    cy.get(FILTERS_DROPDOWN)
+      .contains('Clusters impacted')
+      .then((element) => {
```
gkarat, line 294: Why do you call `then` when you can just directly call `click`? 
```suggestion
      .click({ force: true });
```

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -282,6 +282,45 @@ describe('successful non-empty recommendations list table', () => {
...
+        cy.wrap(element);
+        element[0].click();
+      });
+    cy.get(FILTER_TOGGLE).then((element) => {
+      cy.wrap(element);
+      element[0].click();
+    });
+    cy.get('#pf-random-id-72-false').check({ force: true });
```
gkarat, line 302: This selector is quite unreliable. For example, if you rerun this test, it will fail for you (and did fail for me). You must not use randomly generated class names or ids as they are different with each build.

Fewwy:
@gkarat PTAL

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -282,6 +282,45 @@ describe('successful non-empty recommendations list table', () => {
...
+    cy.get('.pf-c-chip-group__list-item').contains('1 or more');
+
+    cy.get(RECS_LIST_TABLE).find('button[class=pf-c-dropdown__toggle]').click();
+    cy.get(FILTERS_DROPDOWN)
+      .contains('Status')
+      .then((element) => {
```
gkarat, line 308: the same for `then` and `click` stuff

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -282,6 +282,45 @@ describe('successful non-empty recommendations list table', () => {
...
+        cy.wrap(element);
+        element[0].click();
+      });
+    cy.get(FILTER_TOGGLE).then((element) => {
```
gkarat, line 312: the same for `then` and `click` stuff

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -282,6 +282,45 @@ describe('successful non-empty recommendations list table', () => {
...
+      cy.wrap(element);
+      element[0].click();
+    });
+    cy.get('button[class=pf-c-select__menu-item]')
+      .contains('All')
+      .then((element) => {
```
gkarat, line 318: the same for `then` and `click` stuff

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -282,6 +282,45 @@ describe('successful non-empty recommendations list table', () => {
...
+        cy.wrap(element);
+        element[0].click();
+      });
+    cy.get('.pf-c-chip-group__list-item').contains('1 or more');
```
gkarat, line 322: Maybe better check whether both filter groups are present instead of checking just one (Impacting)? In other words, at the end of the scenario, you test whether no groups/filters are gone.

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -282,6 +282,45 @@ describe('successful non-empty recommendations list table', () => {
       .find('td[data-label=Category]')
       .should('contain', 'Service Availability');
   });
+
+  it('the Impacted filters work correctly', () => {
+    const RECS_LIST_TABLE = 'div[id=recs-list-table]';
+    const FILTERS_DROPDOWN = 'ul[class=pf-c-dropdown__menu]';
```
gkarat, line 288: Please move these selectors definition to the rest of selector (line 11)

gkarat (changes requested):

Fewwy:
@gkarat PTAL

gkarat (approved):

PR #86, opened 2022-01-11
Fixing the N/A is not sortable in the Clusters page

Fewwy:
Solving the issue with an empty string in cluster.last_checked_at


gkarat (approved):
Good job! Approving this.

However, (rather general advice), please, next time, try not to modify/affect other tests when writing a new one. Sometimes, it can lead to unnecessary debugging and refactor. Most probably, you will always want to create another mock data set if the existing one does not suit your purposes.

PR #87, opened 2022-01-11
Compact Labels in Recommendations Table

Fewwy:
Added props isCompact to the labels in Recommendations List Table
To update the label total risk in the expanded rule we need to create a separate patch for the rhel advisor
Asked Martin about <InsightsLabel /> props and he told me that there is a bug so currently, we need to pass rest={{ isCompact: true }} instead isCompact="true" to get the result we need.

image

### DIFF
--- a/src/Components/Labels/CategoryLabel.js
+++ b/src/Components/Labels/CategoryLabel.js
@@ -27,14 +27,14 @@ const CategoryLabel = ({ tags }) => {
   const intl = useIntl();
 
   return (
-    <LabelGroup numLabels={4}>
+    <LabelGroup numLabels={1} isCompact>
       {extractCategories(tags).map((tag, key) => (
         <Label
           key={key}
           icon={CATEGORY_ICONS[tag]}
           variant="outline"
           color="blue"
-          isCompact="true"
+          isCompact
         >
           {intl.formatMessage(messages[camelCase(tag)])}
         </Label>
--- a/src/Components/Labels/CategoryLabel.js
+++ b/src/Components/Labels/CategoryLabel.js
@@ -34,6 +34,7 @@ const CategoryLabel = ({ tags }) => {
           icon={CATEGORY_ICONS[tag]}
           variant="outline"
           color="blue"
+          isCompact="true"
         >
           {intl.formatMessage(messages[camelCase(tag)])}
         </Label>
--- a/src/Components/RecsListTable/RecsListTable.js
+++ b/src/Components/RecsListTable/RecsListTable.js
@@ -104,7 +104,7 @@ const RecsListTable = ({ query }) => {
                     {' '}
                     {value?.description || value?.rule_id}{' '}
                   </Link>
-                  <RuleLabels rule={value} />
+                  <RuleLabels rule={value} isCompact="true" />
                 </span>
               ),
             },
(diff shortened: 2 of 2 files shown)


Fewwy:
@gkarat After discussion with Marie Doruskova it became clear that RHEL Advisor team will fix styling on their end. She advised us to merge this patch as it is and wait for them to change labels for https://github.com/RedHatInsights/insights-advisor-frontend/blob/master/src/PresentationalComponents/RuleDetails/RuleDetails.js

src/Components/Labels/CategoryLabel.js
```diff
@@ -34,6 +34,7 @@ const CategoryLabel = ({ tags }) => {
           icon={CATEGORY_ICONS[tag]}
           variant="outline"
           color="blue"
+          isCompact="true"
```
gkarat, line 37:
```suggestion
          isCompact
```

gkarat: ... also, in your version, you are passing the string with the value "true", but not a boolean. If you want to pass a boolean, you would need to use `isCompact={true}` or just `isCompact` instead.

gkarat (changes requested):

src/Components/Labels/CategoryLabel.js
Fewwy: True! Thanks

src/Components/Labels/CategoryLabel.js
```diff
@@ -34,6 +34,7 @@ const CategoryLabel = ({ tags }) => {
...
         >
           {intl.formatMessage(messages[camelCase(tag)])}
         </Label>
```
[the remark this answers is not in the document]
Fewwy: Yep, that's correct + I had to add the isCompact to the LabelGroup

src/Components/RecsListTable/RecsListTable.js
```diff
@@ -104,7 +104,7 @@ const RecsListTable = ({ query }) => {
                     {' '}
                     {value?.description || value?.rule_id}{' '}
                   </Link>
-                  <RuleLabels rule={value} />
+                  <RuleLabels rule={value} isCompact="true" />
```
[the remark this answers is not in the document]
Fewwy: Yep, you are correct

src/Components/Labels/CategoryLabel.js
```diff
@@ -27,13 +27,14 @@ const CategoryLabel = ({ tags }) => {
...
   return (
-    <LabelGroup numLabels={4}>
+    <LabelGroup numLabels={1} isCompact>
       {extractCategories(tags).map((tag, key) => (
         <Label
           key={key}
           icon={CATEGORY_ICONS[tag]}
           variant="outline"
           color="blue"
+          isCompact
```
gkarat, line 37: Looks unnecessary. If you define `isCompact` for the `LabelGroup`, there is no need to add the same parameter to each of the nested labels: https://www.patternfly.org/v4/components/label-group/.

gkarat (changes requested):

Fewwy:
@gkarat PTAL

src/Components/Labels/CategoryLabel.js
gkarat: okk, checked it on my instance and I have the same issue as yours. 👍🏼  It seems like a bug or smth since their documentation states that the labels must become compact once the parameter is provided for the `LabelGroup`... created https://github.com/patternfly/patternfly-react/issues/6815

gkarat (approved):

PR #88, opened 2022-01-12
Affected clusters

ikerreyes:
Proposal for enhancement of AffectedClusterTable:

- check data is as expected
- use default row number as variable
- loop through pagination options
- loop through different search options
- check "Clear filters" work
- iterate over pages


src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -15,6 +16,10 @@ const PAGINATION_MENU =
   'div[data-ouia-component-type="PF4/PaginationOptionsMenu"]';
 const TOOLBAR_CONTENT = '.pf-c-toolbar__content';
 const TOOLBAR = '.pf-c-toolbar';
+const DEFAULT_ROW_COUNT = 20;
```
gkarat, line 19: Nit-pick: if you want to keep this value in sync with the source code, better to import the default value from https://github.com/RedHatInsights/ocp-advisor-frontend/blob/master/src/Services/Filters.js#L4
```
import { AFFECTED_CLUSTERS_INITIAL_STATE } from '../../../Services/Filters.js'

...

const DEFAULT_ROW_COUNT = AFFECTED_CLUSTERS_INITIAL_STATE.limit;
```

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -15,6 +16,10 @@ const PAGINATION_MENU =
...
+// FIXME is this shared by all tables?
```
gkarat, line 20: The pagination values are implicitly given by the `PrimaryToolbar` component that we import from `@redhat-cloud-services/frontend-components` package and the `Pagination` component that we import from `@patternfly/react-core`. We do not override them in any of the tables, which means yes, the values are shared across all tables. I think this is OK to leave it here, in spec, for now.

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -27,6 +32,51 @@ Cypress.Commands.add('getToggleCheckboxText', () =>
   cy.get(AFFECTED_LIST_TABLE).find(TOOLBAR).find('#toggle-checkbox-text')
 );
 
+function filterData(text = '') {
+  // FIXME: is this the right way to use loadash?
```
gkarat, line 36: I didn't know that Cypress exposes lodash :-D I think yes, this is another legit way to use lodash. 👍🏼 

Just in case, you also can import lodash functions directly here in any of the JS files (modules) since `ocp-advisor` has a lodash dependency and imported from `node_modules`.

e.g.
```
import camelCase from 'lodash/camelCase'
...
const abc = camelCase(...);

File: src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js [hunk 4]
```diff
@@ -27,6 +32,51 @@ Cypress.Commands.add('getToggleCheckboxText', () =>
...
+  return Cypress._.filter(props['enabled'], (it) =>
+    (it?.cluster_name || it.cluster).toLowerCase().includes(text.toLowerCase())
+  );
+}
+
+// FIXME improve syntax
+// FIXME move to utils module
+function itemsPerPage() {
+  let items = filterData().length;
+  const array = new Array();
```
gkarat, line 46:
```suggestion
  const array = [];
```

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -7,6 +7,7 @@ import { AffectedClustersTable } from './AffectedClustersTable';
 import props from '../../../cypress/fixtures/AffectedClustersTable/data.json';
 import { Intl } from '../../Utilities/intlHelper';
 import getStore from '../../Store';
+import { array } from 'prop-types';
```
gkarat, line 10: It seems like imported by accident. Please, remove it since it is not used anywhere

gkarat: This is actually the cause for the lint failure in your build :-D

gkarat (changes requested):

gkarat (approved):

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
ikerreyes: Nice, but I think we should do that only in case we are the ones deciding the actual value. If it is someone that will just say, "hey, use this value" rather than changing it, then I think it is better to have it in the test rather than importing. Is this value up to us, or are UX designers or PMs deciding it?

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
ikerreyes: Thanks

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
ikerreyes: Good catch! Thank you

ikerreyes:
@gkarat What do you think of using ``TABLE`` instead of ``AFFECTED_LIST_TABLE`` for: 
```
const AFFECTED_LIST_TABLE = 'div[id=affected-list-table]';
```
given that we only have one table in the component?

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
gkarat: Oh, yes, this makes sense, thank you, let's keep it as it is. I think this is a decision of UXs.

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
gkarat: Unfortunately, it is not possible because they do not have an `export` statement for the Pagination values https://github.com/RedHatInsights/frontend-components/blob/master/packages/components/src/Pagination/Pagination.js#L10. Thus, you can't import and share this object. Adding `export`, however, must allow doing this. Do you want me to create a patch?

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
gkarat: Yes, it is a legit way of importing lodash. However, this is not a good import practice, since the whole library of lodash is loaded into the build. You have to to use the following notation instead:
```
import camelCase from 'lodash/camelCase';
...
const foobar = camelCase(...);
```

gkarat:
> @gkarat What do you think of using `TABLE` instead of `AFFECTED_LIST_TABLE` for:
> 
> ```
> const AFFECTED_LIST_TABLE = 'div[id=affected-list-table]';
> ```
> 
> given that we only have one table in the component?

yep, good by me 👍🏼

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
ikerreyes: Great, thank you!

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
ikerreyes: OK, I see. Thanks for the clarification. Then, given than all tables (with pagination) will be using the same values, is there a way in our repo to put `const PAGINATION_VALUES = [10, 20, 50, 100];` that we can then import from our cypress test files? Just to avoid copying that in all cypress test files.

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
ikerreyes: It seems that I need to use
```
import camelCase from 'lodash';
...
const foobar = camelCase(...);
```
otherwise it complains. Will this be good enough or am I doing something wrong?

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
gkarat: yep, you can simply define the array and export it. this is how it goes if you have two modules (js files) A and B:
A.js: 
```
...
export const PAGINATION_VALUES = [10, 20, 50, 100];
...
```
B.js:
```
import { PAGINATION_VALUES } from './A.js';
...
(use PAGINATION_VALUES here)
```

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
gkarat: hm, in fact, it must not complain. however, given that this is rather a nit-pick, feel free to use this last option that you used. i will take a look later on if I can fix it. @ikerreyes

gkarat (approved):

gkarat:
overall, LGTM. just ping me, and we can merge it

PR #89, opened 2022-01-12
Fix corner case when cluster list receive 0 clusters

Fewwy:
New empty state for the Clusters table that will be rendered when the clusters.length === 0 
image

### DIFF
--- a/src/Components/ClustersListTable/ClustersListTable.js
+++ b/src/Components/ClustersListTable/ClustersListTable.js
@@ -11,6 +11,8 @@ import {
   TableVariant,
 } from '@patternfly/react-table';
 import { Card, CardBody } from '@patternfly/react-core/dist/js/components/Card';
+import { Bullseye } from '@patternfly/react-core/dist/js/layouts/Bullseye';
+import { Spinner } from '@patternfly/react-core/dist/js/components/Spinner';
 import {
   Pagination,
   PaginationVariant,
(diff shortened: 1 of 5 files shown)


gkarat (commented):
I believe we need to use another message because suggesting a user to edit filters for the cluster that doesn't have any recommendations is misleading. Please, copy-paste the message from the RHEL's implementation.

image

Fewwy:
Created a new empty state component that will render when the length of the clusters array is 0
image

src/Components/ClustersListTable/ClustersListTable.js
```diff
@@ -171,78 +175,88 @@ const ClustersListTable = ({
...
-        }}
-        filterConfig={{ items: filterConfigItems }}
-        activeFiltersConfig={activeFiltersConfig}
-      />
-      {(isUninitialized || isFetching) && <Loading />}
-      {isError && (
-        <Card ouiaId="error-state">
+    <>
+      {clusters.length === 0 ? (
+        <Card ouiaId="empty-state">
```
gkarat, line 180: Please, remove the card since there is no card in the mock-up.

gkarat (changes requested):

src/Components/ClustersListTable/ClustersListTable.js
Fewwy: Fixed it. Updated the screenshot on top of PR

src/Components/ClustersListTable/ClustersListTable.js
```diff
@@ -171,78 +175,88 @@ const ClustersListTable = ({
...
+                  index: filters.sortIndex,
+                  direction: filters.sortDirection,
+                }}
+                onSort={onSort}
+                isStickyHeader
+              >
+                <TableHeader />
+                <TableBody />
+              </Table>
+              {clusters.length > 0 && filteredRows.length === 0 && (
```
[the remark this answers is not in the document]
Fewwy: True, that's unnecessary because it assumed that the `clusters.length` is more than 0 if we got this line of code

gkarat (approved):

PR #91, opened 2022-01-18
ClusterRules tests improvements.

ikerreyes:
- Add data checks
 - Check total risk  
 - Loop through sorting options
 - Add tests for filters

Data has been moved to different location as for other test files

For tests for filters, I have explored writing all possible combinations (~900) but takes too long. Finally, I think the best approach is to test each filter in isolation (e.g. only change risk) and manually write some combinations.

There are some TODOs that either need a fix in the code or will be processed as a follow up.


src/Components/ClusterRules/ClusterRules.spec.ct.js
```diff
@@ -3,9 +3,54 @@ import { mount } from '@cypress/react';
 import { IntlProvider } from 'react-intl';
 import { MemoryRouter, Route } from 'react-router-dom';
 import { Provider } from 'react-redux';
+import filter from 'lodash/filter';
+import map from 'lodash/map';
+import uniq from 'lodash/uniq';
+import sortBy from 'lodash/sortBy';
+import reverse from 'lodash/reverse';
+import flatMap from 'lodash/flatMap';
```
gkarat, line 11: hm, here, it seems it worked to you right? :D did you have any complaints from cypress while running this?

ikerreyes: No complaints, it is working fine. Thanks for the advice

src/Components/ClusterRules/ClusterRules.spec.ct.js
```diff
@@ -26,54 +67,393 @@ describe('cluster rules table', () => {
...
-            >
-              <Route path="/clusters/:clusterId">
-                <ClusterRules reports={reports} />
-              </Route>
-            </MemoryRouter>
-          </Provider>
-        </IntlProvider>
-      );
-    });
+    mount(
```
gkarat, line 70: If we use the same components in the `mount`, we can extract it from this `beforeEach` block to the separate `before` block. This way, you wouldn't need to copy/paste the same component for each `describe` block. The same works for interceptors (and actually everything which all `describe` scenarios have in common.

src/Components/ClusterRules/ClusterRules.spec.ct.js
```diff
@@ -26,54 +67,393 @@ describe('cluster rules table', () => {
...
+  it('cannot add filters', () => {
+    cy.get('input[data-ouia-component-type="PF4/TextInput"]').type('some text');
+    cy.get(CHIP_GROUP).should('not.exist');
+    cy.get('div.ins-c-conditional-filter')
+      .find('button[data-ouia-component-type="PF4/DropdownToggle"]')
+      .should('be.disabled');
+  });
+
+  it('renders no recommendation message', () => {
+    // FIXME makes sense to text for both strings in the card or as both are bundled together with one is enough?
```
gkarat, line 203: I think one is good enough

gkarat (approved):
such a good job is done! 👍🏼 thank you. superficially, LGTM.  Feel free to merge it anytime soon

src/Components/ClusterRules/ClusterRules.spec.ct.js
ikerreyes: We are using the same component, but `beforeEach`  gives us some nice features, as resting filters and sorting without requiring a cleanup after the test

src/Components/ClusterRules/ClusterRules.spec.ct.js
```diff
@@ -26,54 +141,264 @@ describe('cluster rules table', () => {
...
+  Object.entries({
+    description: 'Description',
+    created_at: 'Added',
+    total_risk: 'Total risk',
+  }).forEach(([category, label]) => {
+    ['ascending', 'descending'].forEach((order) => {
+      it(`sort ${order} by ${label}`, () => {
+        const col = `td[data-label="${label}"]`;
+        const header = `th[data-label="${label}"]`;
+        cy.get(col).should('have.length', 6);
```
ikerreyes, line 205: 6 should not be hardcoded

gkarat (approved):
Superficially, LGTM 👍🏼 Could you please resolve the conflicts, so that we can merge it?

gkarat:
thanks, looks like the conflicts are cleaned up 👍🏼

PR #93, opened 2022-01-19
Fixed a style in the recommendation table name column and single recommendation page

Fewwy:
The style for the label in the name column was missing the isCompact props
image


gkarat (changes requested):
Notice that the RuleLabels is used across other pages. Your change makes it compact everywhere (e.g., on the single recommendation page). Please, make this value parametrizable with the default set to full, not compact.

gkarat (approved):
given that we now need to turn all labels into compact, this LGTM, thank you

PR #94, opened 2022-01-24
Cluster recs: support "first" query parameter

gkarat:
The patch enables the usage of the "first" query parameter with the rule id as a valid value. If given, the list will show the requested rule as the first in hitting recommendations for this cluster.

- [x] Single cluster page: accept the `first` query parameters
- [x] Single recommendation page: change the link to each of the clusters


gkarat:
@Fewwy, PTAL 👁️‍🗨️

Fewwy:
@gkarat LGTM

PR #96, opened 2022-01-24
Proposal for views

ikerreyes:
This is a proposal to implement "views" in a similar way as done in the [openshift console repo](https://github.com/openshift/console/tree/22c6951efe7c4bca87f3f934063b9f4dcb0a4058/frontend/packages/integration-tests-cypress/views). The idea is to have a set of reusable "views" with methods that can be called later in the different test.

The view is a simple object enclosing a set of functions or nested object.

In this PR, I have create a view for a "table" that includes the toolbar and the table itself. I have also modified the tests for the AffectedCluster to see how they look like when using the view.

This proposal has a very clear issue: it cannot easily receive parameters to enclose nested calls. E.g. it would be better that it receives an id for the outer context, and then all nested cy.get/cy.find calls are done using that context


ikerreyes:
@quarckster @gkarat @Fewwy please let me know your thoughts about this proposal and whether you think we should try to go in this direction or is just going to make our lives more complex. Thank you

Fewwy:
@ikerreyes LGTM 
If we can use it both for unit tests and integration tests - that's good and will make code more DRYer

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -8,26 +8,20 @@ import { AffectedClustersTable } from './AffectedClustersTable';
...
 const SEARCH_ITEMS = ['ff', 'CUSTOM', 'Foobar', 'Not existing cluster'];
 
-// actions
-Cypress.Commands.add('countRows', (count) => {
-  cy.get(TABLE).find(ROW_GROUP).children().should('have.length', count);
-});
-Cypress.Commands.add('getToggleCheckboxText', () =>
-  cy.get(TABLE).find(TOOLBAR).find('#toggle-checkbox-text')
-);
+view.getToggleCheckboxText = () => {
```
gkarat, line 22: Can't this function be enclosed right in `filterableTable.js`?

gkarat (commented):
Thanks a lot for the PR 👍🏼. TBH, I like the idea, and I consider it a logical continuation of our tests. Since we utilize many table components, this is undoubtedly the first place where we can identify many similarities and, thus, share the same set of testing utilities across all table tests.

Can you point out the concrete obstacles you faced with enclosing nested calls? From what I see, you have actually succeeded in it. I don't really expect any issues there since all the class names within table components are identical. We only put IDs to the outer table component...

ikerreyes:
@gkarat The issues I refer about nested calls are the fact that you receive the ID only once and then any call to one of the elements uses that id as enclosing. Basically it is what I try to address in #97 so we can discuss there.

cypress/views/filtrableTable.js
```diff
@@ -0,0 +1,79 @@
...
+          1
+        );
+      })
+      .should('have.length', 1);
+  },
+  rows: () => {
+    return cy
+      .get('table tbody[role=rowgroup]')
+      .find('[data-ouia-component-type="PF4/TableRow"]');
+  },
```
gkarat, line 27: Nit-pick: function closures with the single `return` statements or single expression can omit curly brackets. E.g., in this case:
```suggestion
  rows: () => cy
      .get('table tbody[role=rowgroup]')
      .find('[data-ouia-component-type="PF4/TableRow"]'),
```

(from the code below this can be applied for the most of the functions)

gkarat (approved):
Just a few nit-picks, otherwise the patch looks great 🔥  Thanks again for working on it. Ping me once it is ready to merge please

cypress/views/filtrableTable.js
ikerreyes: yes, thanks. Should this also apply to functions without return statement? E.g. 

```
checkRowCounts: (n) => {
    cy.get('table tbody[role=rowgroup]')
      .find('[data-ouia-component-type="PF4/TableRow"]')
      .should('have.length', n);
  }
```

gkarat: @ikerreyes, well, this is actually the case when two approaches **can** lead to different results. When the JS function doesn't have a `return` statement, then this function returns `undefined`. However, once you get rid of curly brackets, the return statement will be the result of the expression (in your case, cy. ... .should(...)). If the result of the function is not consumed anywhere, this is safe to change.

ikerreyes: Thanks for the explanation. And would you prefer to change?

gkarat: @ikerreyes, in this case it is safe, I would prefer changing it

PR #97, opened 2022-01-24
Function views

ikerreyes:
Proposal to enhance views proposed in #96 with option to get properties that can be called with this.

The proposal includes to have a `build` function for all views that receives the parameters for the assign the values to the object, so they can be called with `this.X`.

Maybe this proposal is an overkill feature that is not needed.

### DIFF
--- a/cypress/views/filtrableTable.js
+++ b/cypress/views/filtrableTable.js
@@ -1,0 +1,71 @@
+const TOOLBAR = 'div[id="ins-primary-data-toolbar"]';
+const PAGINATION = 'div[data-ouia-component-type="PF4/Pagination"]';
+const PAGINATION_MENU = `${TOOLBAR} ${PAGINATION} div[data-ouia-component-type="PF4/PaginationOptionsMenu"]`;
+const PAGINATION_NEXT = `${TOOLBAR} ${PAGINATION} button[data-action="next"]`;
+const CHIPS = `${TOOLBAR} div[data-ouia-component-type="PF4/ChipGroup"]`;
+const EMPTY_STATE = 'table .pf-c-empty-state';
+
+const filtrableTable = {
+  isDisplayed: (id) => {
+    cy.get(`div[id=${id}]`)
+      .within(($div) => {
+        cy.get(TOOLBAR).should('have.length', 1);
+        cy.get('table').should('have.length', 1);
+        cy.get('div[data-ouia-component-type="RHI/TableToolbar"]').should(
+          'have.length',
+          1
+        );
+      })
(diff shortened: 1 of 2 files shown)


cypress/views/filtrableTable.js
```diff
@@ -0,0 +1,79 @@
...
+const PAGINATION = 'div[data-ouia-component-type="PF4/Pagination"]';
+const PAGINATION_MENU = `${TOOLBAR} ${PAGINATION} div[data-ouia-component-type="PF4/PaginationOptionsMenu"]`;
+const PAGINATION_NEXT = `${TOOLBAR} ${PAGINATION} button[data-action="next"]`;
+const CHIPS = `${TOOLBAR} div[data-ouia-component-type="PF4/ChipGroup"]`;
+const EMPTY_STATE = 'table .pf-c-empty-state';
+
+const filtrableTable = {
+  id: null, // to be filled by build method
+  isDisplayed: function () {
+    // FIXME makes sense to wrap all calls within a get(`div[id=${this.id}]`)?
```
gkarat, line 11: Yes, that makes sense since you start from `cy.root()` only once and then work only within the component that you fetched with `get`.

Another way is to use `find` chained after `get`. However, it limits you only to one call. The `within` is the best fit in your case. 

https://docs.cypress.io/api/commands/get#Get-in-within

gkarat (approved):
Good idea with `id` property and usage of `this`. I think this is quite an easy way to share the context within a view. LGTM

cypress/views/filtrableTable.js
ikerreyes: Thank you very much. And expanding the scope of this idea, should we wrap all `cy.get` / `cy.find` calls made by any function in this module under the root component with its ID?

E.g. this 
```
headers: () => {
    return cy.get('table').find('th');
  },
```
will became something like 
```
headers: () => {
    return cy.get(`div[id=${this.id}]`).find('table').find('th');
  },
```

I consider this as an overkill as we know there is only one `div[id=${this.id}]` wrapping everything and it will probably just slow down the tests, but want to know your opinion on this.

ikerreyes:
@gkarat If you do not mind, I would prefer to first merge #96 and then this one. Let's wait for Dima's review and then we proceed to merge them all.

cypress/views/filtrableTable.js
gkarat: @ikerreyes, I guess the first option is better. I see only one case when this could break: having two or more tables in one testing canvas. However, this is very unlikely and therefore I wouldn't really care for now 😅

PR #98, opened 2022-01-24
Proposal: cypress command to get/find elements by their OUIAId

ikerreyes:


cypress/support/commands.js
```diff
@@ -23,3 +23,18 @@
...
+  { prevSubject: 'optional' },
+  (subject, item) => {
+    const attr = `[data-ouia-component-id="${item}"]`;
+    let result;
+    if (subject) {
+      result = cy.wrap(subject).find(attr);
+    } else {
+      result = cy.get(attr);
+    }
+    return result;
```
gkarat, line 38: Nit-pick: the whole block can be rewritten as
```suggestion
    return subject ? cy.wrap(subject).find(attr) : cy.get(attr);
```

gkarat (approved):
LGTM, thank you. Do you want to merge it now?

cypress/support/commands.js
ikerreyes: Good suggestion, way more clear. Thank you

PR #101, opened 2022-01-31
WIP: Ccxdev 6436 affected clusters last seen

Fewwy:


Fewwy:
I created a new branch with a fix after I found it
That way it will be much clearer and easy to read PR.

PR #102, opened 2022-01-31
Proposal: Views as classes

ikerreyes:
Extend the concept of Views from pure object to "locatable" Widgets based on class syntax. Those classes can be converted into cypress objects using their locators with `instance.locate` or `cy.locate(instance)`.

This proposal includes a "locate" command that works as find or get (depending of whether it is preceded by other command) and for any instance of Widget, it resolves it calling its "locate" method. It also includes a demo of how new components can be defined.

Cons:
- need to use `locate` to resolve elements
- need to use `new` for every nested object, and pass `this` to be passed to be able to reference parent
- cannot define nested classes in-place

Pros:
- easier to extend syntax. New widgets can be created for OUIA components, e.g. `Button()` can create a PF4 button using the OUIA attributes


Fewwy:
@ikerreyes Great idea! Would like to talk with you tomorrow and maybe document a couple of things

gkarat (commented):
Superficially, in this state, lgtm 👍🏼 as we already talked before, please, since it starts to mix up the cypress API (get, find) with your custom version (locate), consider supplying the code with more comments or documentation.

PR #104, opened 2022-02-02
Removed dependency of view

ikerreyes:
Given that usage of views has been postponed it looks weird to mix both syntax.
Thus, we have removed the dependency of fiterableTable

I have not removed the views folder so it does not interfere with #91 Once both are merged, we can remove that dependency.


gkarat (approved):

Fewwy:
@ikerreyes Looks good to me!

quarckster (approved):

PR #106, opened 2022-02-06
Query parameters Tests

Fewwy:
Wrote 2 tests that check the additional query parameters for the Recs list page 
Added fixtures to the tests for single Cluster page
Created new tests to test the query parameters
Improved previous tests we had for the single Cluster page
image

### DIFF
--- a/src/Components/Cluster/Cluster.spec.ct.js
+++ b/src/Components/Cluster/Cluster.spec.ct.js
@@ -7,15 +7,36 @@ import { Cluster } from './Cluster';
 import { Provider } from 'react-redux';
 import getStore from '../../Store';
 import '@patternfly/patternfly/patternfly.scss';
+import singleClusterPageReport from '../../../cypress/fixtures/Cluster/report.json';
 
 describe('cluster page', () => {
   // selectors
   const CLUSTER_HEADER = '#cluster-header';
   const BREADCRUMBS = 'nav[class=pf-c-breadcrumb]';
   const RULES_TABLE = '#cluster-recs-list-table';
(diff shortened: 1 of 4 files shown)


Fewwy:
@quarckster Ptal

quarckster (approved):
LGTM. Thanks!

src/Components/Cluster/Cluster.spec.ct.js
```diff
@@ -55,7 +77,10 @@ describe('cluster page', () => {
     cy.get(CLUSTER_HEADER).should('have.length', 1);
     // renders table component
     cy.get(RULES_TABLE).should('have.length', 1);
+    // test how many rows were rendered
+    cy.getAllRows().should('have.length', 10);
```
ikerreyes, line 81: You have the data, so it is better to count the number of expected rows from there directly. This way, the data can be expanded without affecting the test

src/Components/Cluster/Cluster.spec.ct.js
```diff
@@ -110,4 +136,46 @@ describe('cluster page', () => {
     cy.get(RULES_TABLE).should('have.length', 0);
     cy.get('.pf-c-empty-state').should('have.length', 1);
   });
+
+  it('adds additional filters passed by the query parameters №1', () => {
+    mount(
+      <MemoryRouter initialEntries={['?total_risk=1&text=foo+bar&category=2']}>
```
ikerreyes, line 142: Just curiosity, why the parameter is a list?

ikerreyes (commented):
Overall looks good to me

PR #108, opened 2022-02-09
Fixed the single cluster page header display name

Fewwy:
Added a query to the smart proxy file that should get the display name for the cluster
attached it to the components
removed the unnecessary AMS query
image
image

### DIFF
--- a/src/Services/SmartProxy.js
+++ b/src/Services/SmartProxy.js
@@ -26,6 +26,11 @@ export const SmartProxyApi = createApi({
     getClusters: builder.query({
       query: () => `v2/clusters`,
     }),
+    getDisplayName: builder.query({
+      query: ({ id, includeDisabled }) =>
+        `v2/cluster/${id}/reports?get_disabled=${includeDisabled}`,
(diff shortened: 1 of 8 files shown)


src/Services/SmartProxy.js
```diff
@@ -26,6 +26,11 @@ export const SmartProxyApi = createApi({
     getClusters: builder.query({
       query: () => `v2/clusters`,
     }),
+    getDisplayName: builder.query({
```
gkarat, line 29: why create an extra query? you can utilize `getClusterById`, since the response already contains the required display name value

gkarat: Yes, I see. But you will make **two** identical requests. And what is a problem with just making a selector to the data already fetched and saved at the redux store?

Fewwy: Hm! That's an idea, thanks

Fewwy: @gkarat PTAL

Hyperkid123 (approved):

PR #110, opened 2022-02-11
Tests for bulk selector in affected clusters table

ikerreyes:
Set of tests for bulk selector.

Probably these test go beyond the scope of the definition of the bulk selector that we do in the code, thus limiting the scope of these tests is open for discussion.


ikerreyes:
Would it make sense to define the function to mount the component (what we do in beforeEach) as a function and divide our tests further into more "describe" sections grouped by what we are testing, e.g. filtering, bulk selector... ?

Fewwy (approved):
@ikerreyes Looks good to me. Need moar tests!

PR #112, opened 2022-02-14
Fixed the cluster name in the modal

Fewwy:
The variable containing cluster_name was updated and required this change

Before
image

After
image


Fewwy:
@RedHatInsights/ui-reviewers Small patch PTAL

Hyperkid123 (approved):

PR #113, opened 2022-02-14
Tables tests

ikerreyes:
Move default variables to module in cypress (`cypress/utils`). This way multiple test can import them.

Add some fake responses from backend on different endpoints:
- /rule/{ruleId}
- /rule/{ruleId}/cluster_detail
- /ack/{ruleId}

Improve tests in AffectedClustersTable:
- check data contains at least on N/A time
- no chips are displayed by default
- sorting tests

ClusterHeader:
- add comments on items to be fixed later

ClustersListTable tests:
- check data
- pagination tests
- use names if available
- sorting

RecsListTable tests:
- minor renames of variables


src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -447,6 +469,65 @@ describe('non-empty successful affected clusters table', () => {
       .eq(0)
       .should('have.text', 'dd2ef343-9131-46f5-8962-290fdfdf2199');
   });
+
+  // TODO fix test: double sorting? keep ordering in single sorting?
+  Object.entries({
+    name: 'Name',
```
gkarat, line 475: probably could be brought to a constant/variable

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -447,6 +469,65 @@ describe('non-empty successful affected clusters table', () => {
...
+    last_checked_at: 'Last seen',
+  }).forEach(([category, label]) => {
+    ['ascending', 'descending'].forEach((order) => {
```
gkarat, line 478: the same here

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -447,6 +469,65 @@ describe('non-empty successful affected clusters table', () => {
...
+      it(`sort ${order} by ${label}`, () => {
+        const col = `td[data-label="${label}"]`;
+        const header = `th[data-label="${label}"]`;
+
+        cy.get(col).should(
+          'have.length',
+          Math.min(DEFAULT_ROW_COUNT, data['enabled'].length)
+        );
+        cy.get(header).find('button').click();
+        // FIXME right way to do the second click?
```
gkarat, line 488: https://docs.cypress.io/api/commands/dblclick ?

src/Components/ClusterHeader/ClusterHeader.spec.ct.js
```diff
@@ -13,6 +13,7 @@ const LAST_SEEN_FIELD = '#cluster-header-last-seen > :nth-child(2)';
 describe('cluster page header', () => {
   let props;
 
+  //FIXME where is clusterId computed?
```
gkarat, line 16: what do you mean? we receive cluster ids in the API response

src/Components/ClusterHeader/ClusterHeader.spec.ct.js
```diff
@@ -43,6 +44,7 @@ describe('cluster page header', () => {
     // check last seen text
     cy.get(LAST_SEEN_FIELD).should('have.text', '24 Jul 2021 14:22 UTC');
   });
+  //FIXME spinner does not affect UUID/last seen?
   it('show spinner when in the loading state', () => {
```
gkarat, line 48: nit-pick: it is actually a skeleton, not spinner

src/Components/ClusterHeader/ClusterHeader.spec.ct.js
```diff
@@ -43,6 +44,7 @@ describe('cluster page header', () => {
     // check last seen text
     cy.get(LAST_SEEN_FIELD).should('have.text', '24 Jul 2021 14:22 UTC');
   });
+  //FIXME spinner does not affect UUID/last seen?
```
gkarat, line 47: uuid is rendered immediately (know from URL). however, last seen and cluster display name are taken from API response. if API request is ongoing, the skeleton rendered for both fields. if suceeds, names and timestamp written. if last seen unavailable, then Unknown rendered. if cluster name empty/unavailable, the cluster id is rendered in the title

src/Components/ClustersListTable/ClustersListTable.spec.ct.js
```diff
@@ -43,21 +156,79 @@ describe('clusters list table', () => {
...
+  it('can change page limit', () => {
+    // FIXME: best way to make the loop
+    cy.wrap(PAGINATION_VALUES).each((el) => {
+      cy.get(TOOLBAR).find(PAGINATION_MENU).find(DROPDOWN_TOGGLE).click();
+      cy.get(TOOLBAR)
+        .find(PAGINATION_MENU)
+        .find('ul[class=pf-c-options-menu__menu]')
+        .find(DROPDOWN_ITEM)
+        .contains(`${el}`)
+        .click({ force: true }); // caused by the css issue
```
gkarat, line 197: wow, I am surpried we still have the issue with the elements overlapping. must be addressed later on

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -482,3 +476,5 @@ describe('Recs list is requested with additional parameters №2', () => {
     getChipGroup('Clusters impacted').contains('.pf-c-chip', '1 or more');
   });
 });
+
+// TODO what will happen on request error?
```
gkarat, line 480: this
image

gkarat (approved):
I didn't dig into each test's details much, but by their descriptions and the superficial overview, this looks good to me! thanks, this is quite a good work and refactor 🏅

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -16,19 +16,23 @@ import {
...
 const TABLE = 'div[id=affected-list-table]';
 const BULK_SELECT = '[data-ouia-component-id="clusters-selector"]';
-const DEFAULT_ROW_COUNT = 20;
-// FIXME is this shared by all tables?
-const PAGINATION_VALUES = [10, 20, 50, 100];
 const SEARCH_ITEMS = ['ff', 'CUSTOM', 'Foobar', 'Not existing cluster'];
+const TABLE_HEADERS = ['Name', 'Last seen'];
 
 function filterData(text = '') {
   // FIXME: is this the right way to use loadash?
```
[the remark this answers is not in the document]
ikerreyes: Nice, thank you. I was not aware, but given that we have lodash available, I think it is easy to use it.

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -70,11 +74,20 @@ describe('test data', () => {
   });
   it('has one enabled clusters with "foobar" in name', () => {
     cy.wrap(filterData('foobar')).its('length').should('be.eq', 1);
+    expect(_.map(SEARCH_ITEMS, (it) => it.toLowerCase())).to.include('foobar');
```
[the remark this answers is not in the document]
ikerreyes: Good catch. Thank you

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
ikerreyes: totally right, thanks

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
ikerreyes: Good idea, thank you. I propose to use 

```_.zipObject(['name', 'last_checked_at'], TABLE_HEADERS)```

Do you think it will make sense to define `['name', 'last_checked_at']` as a constant? It is only used in this test

File: src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js (hunk 3 above)
Comment (ikerreyes) [reply src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js]: Nice, thank you! I missed it

File: src/Components/ClusterHeader/ClusterHeader.spec.ct.js (hunk 4 above)
Comment (ikerreyes) [reply src/Components/ClusterHeader/ClusterHeader.spec.ct.js:16]: I think no endpoint returns "clusterID" as field. Probably it is copied from another field in the API response and then converted in the format used by the object. Where is that conversion done?

File: src/Components/ClusterHeader/ClusterHeader.spec.ct.js (hunk 5 above)
Comment (ikerreyes) [reply src/Components/ClusterHeader/ClusterHeader.spec.ct.js:48]: Thanks for pointing out. I should learn the right terms

File: src/Components/ClusterHeader/ClusterHeader.spec.ct.js (hunk 6 above)
Comment (ikerreyes) [reply src/Components/ClusterHeader/ClusterHeader.spec.ct.js]: Nice trick. Thanks for explaining

File: src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js (hunk 1 above)
Comment (gkarat) [reply src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js]: I see you use `Object.entries` on `zipObject`. Can it be replaced just with `zip` instead? The results will be the same: an array of 2-items-length arrays.

Comment (gkarat) [reply src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js]: If only in this test, then yeah, it doesn't make sense :-)

File: src/Components/ClusterHeader/ClusterHeader.spec.ct.js (hunk 4 above)
Comment (gkarat) [reply src/Components/ClusterHeader/ClusterHeader.spec.ct.js:16]: Oh, gotcha. The `clusterId` property is sent from src/Components/ClusterHeader/index.js, where it is directly taken from URL path part (`match.params.clusterId`). In other words, once users are at advisor/clusters/foobar page, then the `clusterId` has the value `foobar`. And this has nothing to do with API 👍🏼

File: src/Components/ClustersListTable/ClustersListTable.spec.ct.js (hunk 7 above)
Comment (ikerreyes) [reply src/Components/ClustersListTable/ClustersListTable.spec.ct.js]: Actually is not needed. I check and most of them can be removed, so I have done

File: src/Components/RecsListTable/RecsListTable.spec.ct.js (hunk 8 above)
Comment (ikerreyes) [reply src/Components/RecsListTable/RecsListTable.spec.ct.js:469]: Is there any way to test that? Maybe we should create a test for that. What do you think? Or is not needed?

File: src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js (hunk 1 above)
Comment (ikerreyes) [reply src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js]: Great suggestion, way better. Thank you!

File: src/Components/ClusterHeader/ClusterHeader.spec.ct.js (hunk 4 above)
Comment (ikerreyes) [reply src/Components/ClusterHeader/ClusterHeader.spec.ct.js:16]: OK, understood. Thank you!

File: src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js (hunk 10 above)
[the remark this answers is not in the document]
Comment (ikerreyes) [reply src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js]: Replaced with 

```
cy.wrap(_.map(SEARCH_ITEMS, (it) => it.toLowerCase())).should((arr) => {
      expect(arr).to.include('foobar');
    });
```

What do you think?

src/Components/RecsListTable/RecsListTable.spec.ct.js
ikerreyes: Just asking because if I am not mistaken we check the error status in other tables

gkarat: Yes, you can simply put `isError` to the `query` object within props. Also, the same message will appear when `isSuccess` is set to `true`, but `data` has no entries (recommendations). This cannot happen because, in the response, there will be at least the recommendations that do not hit any clusters.

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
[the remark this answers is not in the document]
gkarat: I think it's ok, the was no problem in the code actually. My comment was rather about that the description of the test doesn't match with what you are asserting on the next line (asserting the contents of SEARCH_ITEMS is somewhat different). But again, it's just a very minor thing 😅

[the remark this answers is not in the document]
ikerreyes: You are right, description does match. I will fix that as well

gkarat:
💯 merging it!

PR #116, opened 2022-02-17
Fixed a bug with wrong mapping on ClusterList page

Fewwy:
Fixed a bug with wrong mapping on ClusterList page
Added a test to check if the fix works

image
image

### DIFF
--- a/src/Components/Common/Tables.js
+++ b/src/Components/Common/Tables.js
@@ -84,10 +84,10 @@ export const mapClustersToRows = (clusters) =>
         </Link>
       </span>,
       cluster.total_hit_count,
+      cluster.hits_by_total_risk?.[4] || 0,
       cluster.hits_by_total_risk?.[3] || 0,
       cluster.hits_by_total_risk?.[2] || 0,
       cluster.hits_by_total_risk?.[1] || 0,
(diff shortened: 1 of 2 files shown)


src/Components/Common/Tables.js
```diff
@@ -84,10 +84,10 @@ export const mapClustersToRows = (clusters) =>
         </Link>
       </span>,
       cluster.total_hit_count,
+      cluster.hits_by_total_risk?.[4] || 0,
       cluster.hits_by_total_risk?.[3] || 0,
       cluster.hits_by_total_risk?.[2] || 0,
       cluster.hits_by_total_risk?.[1] || 0,
```
ikerreyes, line 90: This is getting key '1' as string or the first value in the object?

Fewwy: Yes
It goes cluster -> hits_by_total_risk -> ?. check if the index 1/2/3/4 exist if it exist it will pull up the data

ikerreyes: But we should not check index but keys. The response `{ "1": 3, "2": 9, "3": 3, "4": 0 }` is a mapping is does not guarantee order

ikerreyes (approved):

PR #119, opened 2022-02-23
Single cluster page disabled filter

Fewwy:
Added a ternary operator to check if the cluster has rule hits
If the cluster has 0 rule hits filter in the primary toolbar will be disabled
Otherwise, the filter in the primary toolbar will work

image

image

### DIFF
--- a/src/Components/ClusterRules/ClusterRules.js
+++ b/src/Components/ClusterRules/ClusterRules.js
@@ -369,8 +369,7 @@ const ClusterRules = ({ reports }) => {
         expandAll={{ isAllExpanded, onClick: collapseAll }}
         filterConfig={{
           items: filterConfigItems,
-          isDisabled:
-            reports.length === 0 ? filteredRows.length === 0 : undefined,
+          isDisabled: reports.length === 0,
         }}
         pagination={
           <React.Fragment>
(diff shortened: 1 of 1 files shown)


src/Components/ClusterRules/ClusterRules.js
```diff
@@ -369,7 +369,8 @@ const ClusterRules = ({ reports }) => {
         expandAll={{ isAllExpanded, onClick: collapseAll }}
         filterConfig={{
           items: filterConfigItems,
-          isDisabled: filteredRows.length === 0,
+          isDisabled:
+            reports.length === 0 ? filteredRows.length === 0 : undefined,
```
gkarat, line 373: Check for the length of the `filteredRows` is unnecessary: there is no case when we want to disable the toolbar depending on the filtered entities (and this is a thing I confused when I was implementing this piece of code). It needs to depend only on `reports` - the initial array of entities before we filter them. Thus, I suggest the following:
```suggestion
            reports.length === 0;

Fewwy: Yep this is a better variant :) thanks for the comment

gkarat (approved):

PR #121, opened 2022-02-28
Tests for ReportDetails

ikerreyes:


ikerreyes:
Is it possible to test async components? I have not been able to make it render.

gkarat:
This would be tricky. The problem is that the component you are trying to test is a federated module that is actually stored in the RHEL Advisor source code. The component is fetched and rendered in the runtime from c.r.c. CDN.

One of my current goals is to agree with the RHEL Advisor team on how we are going to treat and take care of such federated modules. My proposal will be to keep all the federated modules only in one application ideally. Given how well our tables and other components are covered with tests, I think it is good to keep all the components under the OCP roof. Once we have this agreement, we will be able to implement tests for the remaining components (including ReportDetails).

I hope this doesn't really block you right now. But if so, we can improvise and come up with the tentative solution (contribute to RHEL Advisor repo?).

cc: @Fewwy

PR #128, opened 2022-03-01
Cluster List added a check for the clusters that missing both clustername and external id

Fewwy:
Added a check for the clusters that missing both cluster name and external id
image

### DIFF
--- a/src/Services/SmartProxy.js
+++ b/src/Services/SmartProxy.js
@@ -25,11 +25,11 @@ export const SmartProxyApi = createApi({
     }),
     getClusters: builder.query({
       query: () => `v2/clusters`,
-    }),
-    getDisplayName: builder.query({
-      query: ({ id, includeDisabled }) =>
-        `v2/cluster/${id}/reports?get_disabled=${includeDisabled}`,
-      transformResponse: (response) => response?.report?.meta?.cluster_name,
+      transformResponse: (response) => {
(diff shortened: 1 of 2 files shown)


src/Services/SmartProxy.js
```diff
@@ -25,11 +25,11 @@ export const SmartProxyApi = createApi({
     }),
     getClusters: builder.query({
       query: () => `v2/clusters`,
-    }),
-    getDisplayName: builder.query({
-      query: ({ id, includeDisabled }) =>
-        `v2/cluster/${id}/reports?get_disabled=${includeDisabled}`,
-      transformResponse: (response) => response?.report?.meta?.cluster_name,
+      transformResponse: (response) => {
```
gkarat, line 28: Nit-pick: you can omit the usage of curly brackets here since the function consists only of the return statement.

src/Services/SmartProxy.js
```diff
@@ -25,11 +25,11 @@ export const SmartProxyApi = createApi({
...
+        return {
+          data: response.data.filter((element) => element.cluster_id !== ''),
```
gkarat, line 30: Nit-pick: can be rewritten to
```suggestion
          data: response.data.filter((element) => element.cluster_id),
```

gkarat (approved):

PR #129, opened 2022-03-01
Copy-paste the rule details fed module from RHEL

gkarat:
This is a temporary fix for the CCXDEV-7412: RHEL Advisor does not
provide the latest version of the AdvisorRecommendationDetails federated
module in prod-stable environment. This copy-pastes the latest version
of the component to the OCP Advisor.

Should be reverted once RHEL Advisor is updated.


Fewwy:
@gkarat Tested it locally LGTM

Fewwy (approved):

PR #133, opened 2022-03-02
Clusters list, recommendations list pages: solving the UI inconstistency

Fewwy:
Added a parameter to the clusters page endpoint and recommendation page endpoint
Added a parameter to all smart proxy endpoints to fetch data if the user lost his internet and then got it back

Link to the RTK docs https://redux-toolkit.js.org/rtk-query/usage/cache-behavior#default-cache-behavior

### DIFF
--- a/src/Services/SmartProxy.js
+++ b/src/Services/SmartProxy.js
@@ -7,6 +7,7 @@ export const SmartProxyApi = createApi({
   baseQuery: fetchBaseQuery({
     baseUrl: BASE_URL,
   }),
+  refetchOnReconnect: true,
   endpoints: (builder) => ({
     getClusterById: builder.query({
       query: ({ id, includeDisabled }) =>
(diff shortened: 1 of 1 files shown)


src/Services/SmartProxy.js
```diff
@@ -30,6 +31,7 @@ export const SmartProxyApi = createApi({
           data: response.data.filter((element) => element.cluster_id !== ''),
         };
       },
+      keepUnusedDataFor: 5,
```
gkarat, line 34: Does it make sense to set this value to 0 instead? So that we **always** do re-fetch when the table is mounted again. 

This could reduce potential cludges in the integration tests for @ikerreyes, and we would not need to wait for 5 seconds to have data updated.

gkarat:
The solution is cool, I really appreciate we can get so much advantage of RTK Query. Let's discuss the timeout value and we can merge it 👍🏼

gkarat (approved):

gkarat (approved):
looks good to me, the pages refresh after 5 secs after unmount 👍🏼

PR #135, opened 2022-03-04
Update README.md

gkarat:


gkarat:
Some chapters reordering. The main update is an info of which branches are accountable for which environments with respective links

@ikerreyes, PTAL

README.md
```diff
@@ -4,52 +4,58 @@
 
 ## Running locally
 
-1. `npm install`
+### First time setup
 
-2. Check whether `env` in the `dev.webpack.config.js` is set to the preferred chrome + API environment.
+1. Make sure you have [`Node.js`](https://nodejs.org/en/) and [`npm`](https://www.npmjs.com/) installed.
```
ikerreyes, line 9: maybe it is worth to indicate which node.js version

README.md
```diff
@@ -4,52 +4,58 @@
...
+2. Make sure you are using [Red Hat proxy](http://hdn.corp.redhat.com/proxy.pac).
 
-3. `npm run start:beta`
+### Running a development server
 
-4. Open https://prod.foo.redhat.com:1337/beta/openshifts/insights/advisor (or whatever environment is specified). Please, make sure that OCP Advisor is temporarily not available under **non-beta** environments.
+1. Install dependencies with `npm install`.
+2. Run development server with `npm run start:beta` (`npm run start` for the non-beta release).
```
ikerreyes, line 15: For the development server, it does not really care much about beta and not-beta right? It will just be a difference in  `/beta` in the URL, right? If so, I would propose to set the non-beta as default (for simplicity). Or are there other implications?

ikerreyes (approved):
Nice, thank you very much.

README.md
gkarat: It actually do cares when, from the local development server, you browse other applications (it will serve beta builds if you run under `*-beta` env, and stable builds under `*-stable`). Also, I came across one bug while running locally another app (OCM): while running with beta release, if you browse the app under the stable URL, some of the requests are not passing proxy... This is a long story, but generally want to say that there is a difference

PR #136, opened 2022-03-07
feat: integrate semantic-release

gkarat:
This integrates `semantic-release` with Advisor and adds automatic versioning. 

Commit messages should from now on comply with the Angular commit message conventions: https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines so that a new release can be triggered with each merge to the **master** branch. Types of commits are listed here: https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#type and should be all supported with the configuration introduced by this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,6 @@
   "scripts": {
     "build": "NODE_ENV=production webpack --config config/prod.webpack.config.js",
     "analyze": "NODE_ENV=production webpack --config config/prod.webpack.config.js --env analyze=true",
-    "deploy": "npm-run-all build lint test",
     "lint": "npm-run-all lint:*",
     "lint:js": "eslint config src",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -20,7 +19,8 @@
     "translations": "npm-run-all translations:*",
     "translations:extract": "formatjs extract 'src/**/*.js*' --out-file lang/en.json --id-interpolation-pattern '[sha512:contenthash:base64:6]'",
     "translations:compile": "formatjs compile lang/en.json --out-file compiled-lang/en.json",
-    "verify": "npm-run-all build lint test"
+    "verify": "npm-run-all build lint test",
+    "semantic-release": "semantic-release"
```
ikerreyes, line 23: I think there will be only 1 release command, have you considered just naming it `release`?

PR #177, opened 2022-03-14
fix: Refresh pages and smartproxy fix

Fewwy:


src/Components/RecsListTable/RecsListTable.js
```diff
@@ -78,6 +78,7 @@ const RecsListTable = ({ query }) => {
     setDisplayedRows(
       buildDisplayedRows(filteredRows, filters.sortIndex, filters.sortDirection)
     );
+    setInterval(() => refetch(), 20000);
```
gkarat, line 81: When I ran it locally, every 20 secs, there were two identical /rule requests made instead of only one. I think, the problem is in that you added `setInterval` to a useEffect which is dependent on `filteredRows` and other parameters. Once these params are updated, this will call again for setInterval and extra requests will be stacked with the same interval.

The solution can be simple: create another useEffect with zero dependencies ([]). This will set up the callback only once because the component is mounted only once. Since there are no dependencies, the body of such useEffect will be executed only on mount.

gkarat (changes requested):
refetchOnMountOrArgChange looks good 👍🏼 

However, there is a problem in setInterval, PTAL

src/Components/RecsListTable/RecsListTable.js
Fewwy: Noted and added :)

PR #178, opened 2022-03-15
style(RecsList & Single Cluster page): changed the Added column

Fewwy:
changed the Added column to modified

updated the translations


gkarat (approved):
nice one 👍🏼 one minor thing: I don't think the change refers to the "style" category, but it's rather a feature change. anyway, merging it, lgtm

PR #183, opened 2022-03-16
fix(RecsListTable): Reset filters button behavior update

Fewwy:
According to the task CCXDEV-7614 I updated the "Reset filters" behavior

### DIFF
--- a/src/Components/RecsListTable/RecsListTable.js
+++ b/src/Components/RecsListTable/RecsListTable.js
@@ -19,7 +19,7 @@ import {
   PaginationVariant,
 } from '@patternfly/react-core/dist/js/components/Pagination';
 import { Stack } from '@patternfly/react-core/dist/js/layouts/Stack';
-import { isEqual } from 'lodash';
+import isEqual from 'lodash/isEqual';
 import {
   Tooltip,
   TooltipPosition,
--- a/src/Components/RecsListTable/RecsListTable.js
+++ b/src/Components/RecsListTable/RecsListTable.js
@@ -418,14 +418,23 @@ const RecsListTable = ({ query }) => {
     return pruneFilters(localFilters, FILTER_CATEGORIES);
   };
 
+  console.log(RECS_LIST_INITIAL_STATE);
   const activeFiltersConfig = {
     showDeleteButton: true,
     deleteTitle: intl.formatMessage(messages.resetFilters),
     filters: buildFilterChips(),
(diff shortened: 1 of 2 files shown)


src/Components/RecsListTable/RecsListTable.js
```diff
@@ -418,12 +418,26 @@ const RecsListTable = ({ query }) => {
...
     deleteTitle: intl.formatMessage(messages.resetFilters),
     filters: buildFilterChips(),
     onDelete: (_event, itemsToRemove, isAll) => {
-      if (isAll) {
+      if (
+        itemsToRemove.length === 2 &&
+        itemsToRemove[0].chips[0].name.props.children === '1 or more' &&
+        itemsToRemove[1].chips[0].name.props.children === 'Enabled' &&
+        isAll
+      ) {
```
gkarat, line 432: Your solution does work. But:
1) it can be rewritten in a much shorter way if you simply use RECS_LIST_INITIAL_STATE,
2) if the initial filter set for this table will be changed, then you will have to rewrite also this part of the code. This is bad 👎🏼 
3) your solution doesn't care about translations == you have to use intl.translate to fix it, however, this solution becomes even more unnecessary complicated.

src/Components/RecsListTable/RecsListTable.js
```diff
@@ -418,12 +418,26 @@ const RecsListTable = ({ query }) => {
     return pruneFilters(localFilters, FILTER_CATEGORIES);
   };
 
+  console.log(RECS_LIST_INITIAL_STATE);
```
gkarat, line 421: remove

gkarat (changes requested):

src/Components/RecsListTable/RecsListTable.js
```diff
@@ -418,12 +418,26 @@ const RecsListTable = ({ query }) => {
...
   const activeFiltersConfig = {
+    showDeleteButton: true,
     deleteTitle: intl.formatMessage(messages.resetFilters),
     filters: buildFilterChips(),
     onDelete: (_event, itemsToRemove, isAll) => {
-      if (isAll) {
+      if (
```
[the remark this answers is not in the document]
Fewwy: Должно быть import { isEqual } from 'lodash';
но да, так гораздо более кратко и удобно

gkarat (commented):
This looks good to me, just one minor thing with lodash and we can merge it, thanks

gkarat (approved):

PR #187, opened 2022-03-23
fix(disable modal): bug with multiple selections

gkarat:
This fixes a bug: if a user selects multiple clusters, unchecks the "Disable recommendation for selected clusters" flag, then the recommendation will only be disabled for selected clusters. This is incorrect, and the recommendation should be disabled for all clusters in this scenario.

Tests to be written in https://github.com/RedHatInsights/ocp-advisor-frontend/pull/111.

### DIFF
--- a/src/Components/Modals/DisableRule.js
+++ b/src/Components/Modals/DisableRule.js
@@ -24,9 +24,8 @@ const DisableRule = ({
 }) => {
   const intl = useIntl();
   const [justification, setJustificaton] = useState('');
-  const [singleSystem, setSingleSystem] = useState(
-    host !== undefined || hosts.length > 0
-  );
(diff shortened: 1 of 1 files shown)


src/Components/Modals/DisableRule.js
```diff
@@ -24,9 +24,8 @@ const DisableRule = ({
 }) => {
   const intl = useIntl();
   const [justification, setJustificaton] = useState('');
-  const [singleSystem, setSingleSystem] = useState(
-    host !== undefined || hosts.length > 0
-  );
+  const [singleHost, setSingleHost] = useState(!!host);
```
Fewwy, line 27: Why double negation?
You convert true to true and false to false

Fewwy (approved):
Aside from the double negation LGTM

PR #192, opened 2022-03-24
feat(SingleClusterPage): New dropdown and action button

Fewwy:
1)Added a dropdown to the Cluster Header component
2)Added a dropdown item
3)Set up the redirect to the OCM
screenshots are in the Jira card

### DIFF
--- a/src/Components/ClusterHeader/ClusterHeader.js
+++ b/src/Components/ClusterHeader/ClusterHeader.js
@@ -1,10 +1,11 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { useIntl } from 'react-intl';
 
 import { Grid, GridItem } from '@patternfly/react-core/dist/js/layouts/Grid';
 import { Stack, StackItem } from '@patternfly/react-core/dist/js/layouts/Stack';
 import { Title } from '@patternfly/react-core/dist/js/components/Title';
+import { Dropdown, DropdownToggle, DropdownItem } from '@patternfly/react-core';
 import Skeleton from '@redhat-cloud-services/frontend-components/Skeleton';
 import DateFormat from '@redhat-cloud-services/frontend-components/DateFormat/DateFormat';
 
@@ -12,6 +13,8 @@ import messages from '../../Messages';
 import { OneLineLoader } from '../../Utilities/Loaders';
 
 export const ClusterHeader = ({ clusterId, clusterData }) => {
+  let location = window.location;
+  const [isOpen, setIsOpen] = useState(false);
   const intl = useIntl();
   // subscribe to the cluster data query
   const {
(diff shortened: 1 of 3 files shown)


src/Components/ClusterHeader/ClusterHeader.js
```diff
@@ -20,9 +23,31 @@ export const ClusterHeader = ({ clusterId, clusterData }) => {
...
+  const dropDownToggle = (isOpen) => {
+    setIsOpen(isOpen);
+  };
+
+  const dropDownSelect = () => {
+    setIsOpen(!isOpen);
+  };
+
+  const redirectOCM = (clusterId) => {
+    location.replace(
```
gkarat, line 35: This creates one unwanted behavior: once you click on the button and are redirected to the new page, if you click the browser's "Back", you won't get to the single cluster page in Advisor from where you were navigated to OCM, but to the page before. This is caused by the usage of `replace` which literally replaces your current URL with a new one instead of adding a new one on top of it.
https://developer.mozilla.org/en-US/docs/Web/API/Location/replace

gkarat: Feel free to use direct assignment to `window.location`:
```
const redirectOCM = (clusterId) => {
    window.location =
      window.location.origin +
      (window.location.pathname.includes('beta') ? `/beta` : '') +
      `/openshift/details/${clusterId}`;
  };
```

gkarat (changes requested):

src/Components/ClusterHeader/ClusterHeader.js
```diff
@@ -36,6 +61,20 @@ export const ClusterHeader = ({ clusterId, clusterData }) => {
...
       </GridItem>
+      <GridItem span={4} id="cluster-header-dropdown">
+        <Dropdown
+          position="right"
+          onSelect={dropDownSelect}
+          autoFocus={false}
+          isOpen={isOpen}
+          toggle={
+            <DropdownToggle id="toggle-id-2" onToggle={dropDownToggle}>
+              Action
```
gkarat, line 72: Use `intl` here, please.

src/Components/ClusterHeader/ClusterHeader.js
```diff
@@ -36,6 +61,20 @@ export const ClusterHeader = ({ clusterId, clusterData }) => {
           </Title>
         )}
       </GridItem>
+      <GridItem span={4} id="cluster-header-dropdown">
+        <Dropdown
+          position="right"
+          onSelect={dropDownSelect}
```
gkarat, line 67: A minor suggestion: if it is a one-liner function, you can use an arrow function just right here without creating and referencing a new function.

src/Components/ClusterHeader/ClusterHeader.js
```diff
@@ -36,6 +61,20 @@ export const ClusterHeader = ({ clusterId, clusterData }) => {
...
+          autoFocus={false}
+          isOpen={isOpen}
+          toggle={
+            <DropdownToggle id="toggle-id-2" onToggle={dropDownToggle}>
```
gkarat, line 71: A minor suggestion: if it is a one-liner function, you can use an arrow function just right here without creating and referencing a new function.

gkarat:
Can you please also add tests to this feature? :-)

gkarat (approved):
cool, new buttoonnnn aproachinn

PR #197, opened 2022-03-29
Cluster reports table: refactor and fix the bug with inability to sort the table

gkarat:
* FIX: If a user opens the cluster rules page with the "first" search parameter (== the value of the parameter dictates which rule to show first in the table) and tries to sort by any columns, the table will not react and not sort the reports.
* REFACTOR: remove the part of the table logic from `Cluster` component.


gkarat:
@ikerreyes, tagging you here for visibility :-)

PR #202, opened 2022-03-30
fix(AffectedClustersTable): set default sorting to last seen

Fewwy:
Set default sorting to last seen
Updated 1 failing test
image


gkarat (commented):
lgtm!

PR #205, opened 2022-04-11
Use application basename to react to APP_NAVIGATION event

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-18732


gkarat (approved):
cool, this fixes the issue, thanks for helping with it @Hyperkid123  👍🏼

PR #227, opened 2022-04-19
Import RuleDetails from @redhat-cloud-services/frontend-components-advisor-components

gkarat:
Implements https://issues.redhat.com/browse/CCXDEV-7478.

Use RuleDetails component from @redhat-cloud-services/frontend-components-advisor-components.

Has to be merged after https://github.com/RedHatInsights/frontend-components/pull/1518 is fixed.


src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -420,6 +420,21 @@ describe('successful non-empty recommendations list table', () => {
...
 
+  it('rule content is rendered', () => {
+    // expand all rules
+    cy.get('.pf-c-toolbar__expand-all-icon > svg').click();
+    cy.get(RECS_LIST_TABLE)
+      .find('.pf-c-table__expandable-row.pf-m-expanded')
+      .each((el) => {
+        // contains description
+        cy.wrap(el).contains(
+          'Sed ut perspiciatis unde omnis iste natus error.'
```
ikerreyes, line 431: where is the content defined? Is it coming from one of the already existing fixtures?

ikerreyes (approved):

Fewwy (approved):
I fetched it and check how it works locally and read through your code
Looks good to me :)

src/Components/RecsListTable/RecsListTable.spec.ct.js
gkarat: @ikerreyes, yep, from here https://github.com/RedHatInsights/ocp-advisor-frontend/blob/master/cypress/fixtures/RecsListTable/data.json

ikerreyes: great. Then I think it would be better to check the string from the data itself rather than hardcoding it in the tests, but we can address that later.

gkarat: the test is rather a showcase :) we'll improve it later when focusing more on the testing perspective

PR #228, opened 2022-04-21
Move cypress fixture data files to api folder

ikerreyes:
Cypress data files have been moved to the API folder, so we know were each one is coming. I hope this helps to track api changes easily.

Two have been combined into single file to reduce the chances to make a mistake when api changes.

Fix sorting test when cluster names use names with Capital letters


src/Components/ClusterRules/ClusterRules.spec.ct.js
```diff
@@ -19,6 +19,8 @@ import { applyFilters } from '../../../cypress/utils/filters';
 import { cumulativeCombinations } from '../../../cypress/utils/combine';
 import { CHIP_GROUP, CHIP, ROWS } from '../../../cypress/views/filterableTable';
 
+let data = singleClusterPageReport.report.data;
```
gkarat, line 22: prefer `const` here since the value is not changed farther

src/Components/ClusterRules/ClusterRules.spec.ct.js
```diff
@@ -8,7 +8,7 @@ import _ from 'lodash';
 import getStore from '../../Store';
 import ClusterRules from './ClusterRules';
 import '@patternfly/patternfly/patternfly.scss';
-import data from '../../../cypress/fixtures/ClusterRules/data.json';
+import singleClusterPageReport from '../../../cypress/fixtures/api/insights-results-aggregator/v2/cluster/dcb95bbf-8673-4f3a-a63c-12d4a530aa6f/reports/report-disabled-false.json';
```
gkarat, line 11: assuming that you want to have these routes to be similar to API routes as much as possible, why don't you use `'../../../cypress/fixtures/api/insights-results-aggregator/v2/cluster/dcb95bbf-8673-4f3a-a63c-12d4a530aa6f/reports.json'` instead? 

https://github.com/RedHatInsights/insights-results-smart-proxy/blob/master/server/api/v2/openapi.json#L70

src/Components/Cluster/Cluster.spec.ct.js
```diff
@@ -7,7 +7,7 @@ import { Cluster } from './Cluster';
 import { Provider } from 'react-redux';
 import getStore from '../../Store';
 import '@patternfly/patternfly/patternfly.scss';
-import singleClusterPageReport from '../../../cypress/fixtures/Cluster/report.json';
+import singleClusterPageReport from '../../../cypress/fixtures/api/insights-results-aggregator/v2/cluster/dcb95bbf-8673-4f3a-a63c-12d4a530aa6f/reports/report-disabled-false.json';
```
gkarat, line 10: the same here as for my another comment

src/Components/ClustersListTable/ClustersListTable.spec.ct.js
```diff
@@ -7,7 +7,7 @@ import _ from 'lodash';
 import { Intl } from '../../Utilities/intlHelper';
 import getStore from '../../Store';
 import { ClustersListTable } from './ClustersListTable';
-import props from '../../../cypress/fixtures/ClustersListTable/data.json';
+import props from '../../../cypress/fixtures/api/insights-results-aggregator/v2/clusters/report.json';
```
gkarat, line 10: maybe just
```suggestion
import props from '../../../cypress/fixtures/api/insights-results-aggregator/v2/clusters.json';
```
?

https://github.com/RedHatInsights/insights-results-smart-proxy/blob/master/server/api/v2/openapi.json#L424

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -5,7 +5,7 @@ import { Provider } from 'react-redux';
 
 import { RecsListTable } from './RecsListTable';
 import getStore from '../../Store';
-import data from '../../../cypress/fixtures/RecsListTable/data.json';
+import data from '../../../cypress/fixtures/api/insights-results-aggregator/v2/rule/report.json';
```
gkarat, line 8:
```suggestion
import data from '../../../cypress/fixtures/api/insights-results-aggregator/v2/rule.json';
```
? :D

src/Components/ClusterRules/ClusterRules.spec.ct.js
ikerreyes: because that endpoint is parametrized. It is just to make it clear that it is the response when you request it with `?disabled=false`. But if you prefer to use just `reports`, fine with me.

src/Components/ClustersListTable/ClustersListTable.spec.ct.js
ikerreyes: fine with me, I just wanted to follow the approach that was there. I think this change can be applied to all endpoints. Do you agree to change it everywhere?

src/Components/ClusterRules/ClusterRules.spec.ct.js
ikerreyes: done, thanks

src/Components/ClustersListTable/ClustersListTable.spec.ct.js
ikerreyes: updated, please take a look

gkarat (approved):

PR #235, opened 2022-04-29
Improve url tests

ikerreyes:
RecListTable URL test are improved:

-  looping through parameters
- checking for number of chips

### DIFF
--- a/cypress/utils/globals.js
+++ b/cypress/utils/globals.js
@@ -8,6 +8,17 @@ const CATEGORIES = {
   Performance: ['performance'],
 };
 
+// TODO remove if unused
+// const CATEGORY_MAP = {
+//   'Service Availability': 1,
+//   Performance: 2,
+//   Security: 4,
+//   'Fault Tolerance': 3,
+// };
+
+// const IMPACT = { Low: 1, Medium: 2, High: 3, Critical: 4 };
+// const LIKELIHOOD = { Low: 1, Medium: 2, High: 3, Critical: 4 };
+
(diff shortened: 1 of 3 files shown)


src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -91,34 +94,26 @@ describe('pre-filled url search parameters', () => {
...
+  it('recognizes all parameters', () => {
+    const urlSearchParameters = new URLSearchParams(urlParams);
+    for (const [key, value] of urlSearchParameters) {
+      if (key == 'text') {
+        getChipGroup('Name').contains('.pf-c-chip', value);
+        cy.get('.pf-m-fill > .pf-c-form-control').should('have.value', value);
+      } else {
+        value.split(',').forEach((it) => {
+          const [group, item] = urlParamConvert(key, it);
+          console.log('--', item);
```
gkarat, line 106: remove?

cypress/utils/globals.js
```diff
@@ -8,6 +8,17 @@ const CATEGORIES = {
   Performance: ['performance'],
 };
 
+// TODO remove if unused
```
gkarat, line 11: since it is unused, then can we remove it? 😁

gkarat (approved):
looks good to me. I liked these tests, they are quite thorough and also include the check of the number of chips 👍🏼 just minor comments, but you can basically merge it

cypress/utils/globals.js
ikerreyes: done

PR #239, opened 2022-04-29
Pagination related tests for RecsListTable

ikerreyes:
Used the MR to also add some comments for other improvements in the future

### DIFF
--- a/src/Components/RecsListTable/RecsListTable.spec.ct.js
+++ b/src/Components/RecsListTable/RecsListTable.spec.ct.js
@@ -2,17 +2,31 @@ import React from 'react';
 import { mount } from '@cypress/react';
 import { MemoryRouter } from 'react-router-dom';
 import { Provider } from 'react-redux';
+import _ from 'lodash';
 
 import { RecsListTable } from './RecsListTable';
 import getStore from '../../Store';
-import data from '../../../cypress/fixtures/api/insights-results-aggregator/v2/rule.json';
+import ruleResponse from '../../../cypress/fixtures/api/insights-results-aggregator/v2/rule.json';
 import { Intl } from '../../Utilities/intlHelper';
 import '@patternfly/patternfly/patternfly.scss';
 import {
   TOOLBAR,
   TOOLBAR_FILTER,
   CHIP,
+  PAGINATION,
+  TBODY,
 } from '../../../cypress/utils/components';
+import {
+  DEFAULT_ROW_COUNT,
+  PAGINATION_VALUES,
+} from '../../../cypress/utils/defaults';
+import {
+  checkPaginationTotal,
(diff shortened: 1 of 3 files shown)


src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -2,26 +2,108 @@ import React from 'react';
...
+} from '../../../cypress/utils/pagination';
+import { TOTAL_RISK, CATEGORIES } from '../../../cypress/utils/globals';
 // TODO make more use of ../../../cypress/utils/components
 
 // selectors
 const ROOT = 'div[id=recs-list-table]';
 const ROW = 'tbody[role=rowgroup]'; // FIXME use ROW from components
 const FILTERS_DROPDOWN = 'ul[class=pf-c-dropdown__menu]';
 const FILTER_TOGGLE = 'span[class=pf-c-select__toggle-arrow]';
+const DEFAULT_FILTERS = {
```
gkarat, line 37: Maybe worth to refer to https://github.com/RedHatInsights/ocp-advisor-frontend/blob/master/src/Services/Filters.js#L13

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -209,18 +291,98 @@ describe('successful non-empty recommendations list table', () => {
...
+
+    it('values are expected ones', () => {
+      checkPaginationValues(PAGINATION_VALUES);
+    });
+
+    it('can change page limit', () => {
+      // FIXME: best way to make the loop
+      cy.wrap(PAGINATION_VALUES).each((el) => {
+        changePagination(el).then(() =>
+          // TODO should this be nested. Also the other check below?
```
gkarat, line 343: I believe so, if you don't nest it, then there can be an error/interface not updated as expected after the click. Found this article: https://www.webiny.com/blog/things-to-avoid-when-writing-cypress-tests "If you need to do something after a command has been completed, you can do so with the help of the cy.then command. It will guarantee that only after the previous command finishes, the next will run."

gkarat (approved):
good work :+1 nothing serious, just minor comments

src/Components/RecsListTable/RecsListTable.spec.ct.js
ikerreyes: Makes sense. However, as some of them are not useful for filtering the data, it might require some thinking. I will create a `TODO` and address it separately. Thanks for pointing it out.

src/Components/RecsListTable/RecsListTable.spec.ct.js
ikerreyes: Thanks @gkarat

PR #240, opened 2022-04-29
Move importing PF CSS to cypress support

ikerreyes:
This way it will be autoimported in all test files

Nice finding @gkarat , thank you!


gkarat (approved):
📈

PR #241, opened 2022-04-29
Affected clusters tests improvement

ikerreyes:
- rename data and set default sorting for ease testing
- improve test checking cluster names show names instead of uuids
- improve test checking cluster names are hyperlinks. Test also improved in ClustersListTable tests

### DIFF
--- a/src/Components/ClustersListTable/ClustersListTable.spec.ct.js
+++ b/src/Components/ClustersListTable/ClustersListTable.spec.ct.js
@@ -32,6 +32,7 @@ import {
   changePagination,
 } from '../../../cypress/utils/pagination';
 
+// TODO check if we can rid this of this data and use namedClustersInstead
 const data = props['data'];
 // add property name to clusters
 let namedClusters = _.cloneDeep(data);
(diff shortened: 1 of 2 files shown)


src/Components/ClustersListTable/ClustersListTable.spec.ct.js
```diff
@@ -541,12 +542,16 @@ describe('clusters list table', () => {
...
-        `a[href="/clusters/${namedClustersDefaultSorting[0]['cluster_id']}"]`
-      )
-      .should('have.text', namedClustersDefaultSorting[0]['name']);
+    cy.get(TBODY)
+      .children()
+      .each(($el, index) => {
+        cy.wrap($el)
+          .find('td[data-label=Name]')
+          .find(
+            `a[href="/clusters/${namedClustersDefaultSorting[index]['cluster_id']}"]`
```
gkarat, line 551: does this selector really yield anything? the things is, here, you use `href=` which is an "exact equality" operator. Whereas, I would expect, you would use "*=" here as you did in https://github.com/RedHatInsights/ocp-advisor-frontend/pull/241/files#diff-11510ebd9fcd477bd2c0ab72124a29e1b5a7a8ec9743a472cbc67852ba71b74aR145

ikerreyes: Yes, it seems to be working. In this case the match is exact as it is just a redirection to the cluster, without any other default set up.

image

gkarat: Ah, okay, just checked out that `href` can also equal to a pathname, not the full URL 👍🏼

gkarat (approved):

PR #248, opened 2022-05-02
Fix "Row expanded when sorting is changed"

gkarat:
Fixes https://issues.redhat.com/browse/CCXDEV-7900


ikerreyes (approved):
Please, add a `TODO` note to put the `dblclick` back within a `then` clause

PR #250, opened 2022-05-03
RecsListTable tests improvement

ikerreyes:
- use utils functions
- remove hardcoded values
- improve default tests: pagination, sorting and filtering


gkarat (approved):
superficially, looks good!

PR #251, opened 2022-05-03
fix(ClusterRules): Open rows memorization 

Fewwy:
Fixes https://issues.redhat.com/browse/CCXDEV-7900.

Recording the opened rows for the single cluster page
Fixed the test

The expand collapse behavior:
1)When the user opens the page for the first time - the first row is expanded
2)If he clicks sort after that, only the rows that were expanded will stay that way
if he expands several different rows\all rows before that -> only these rows will stay expanded

### DIFF
--- a/src/Components/ClusterRules/ClusterRules.js
+++ b/src/Components/ClusterRules/ClusterRules.js
@@ -65,12 +65,12 @@ const ClusterRules = ({ cluster }) => {
   const [filteredRows, setFilteredRows] = useState([]);
   const [displayedRows, setDisplayedRows] = useState([]);
   const [isAllExpanded, setIsAllExpanded] = useState(false);
+  const [expandFirst, setExpandFirst] = useState(true);
   const [firstRule, setFirstRule] = useState(''); // show a particular rule first
   const results = filteredRows.length;
   const { search } = useLocation();
   // helps to distinguish the state when the API data received but not yet filtered
   const [rowsFiltered, setRowsFiltered] = useState(false);
-  const [recordExpanded, setRecordExpanded] = useState([]);
   const loadingState = isUninitialized || isFetching || !rowsFiltered;
   const errorState = isError;
   const successState = isSuccess;
@@ -115,21 +115,20 @@ const ClusterRules = ({ cluster }) => {
   const handleOnCollapse = (_e, rowId, isOpen) => {
     const collapseRows = [...displayedRows];
     collapseRows[rowId] = { ...collapseRows[rowId], isOpen };
-    if (!recordExpanded.includes(collapseRows[rowId].rule.rule_id)) {
-      setRecordExpanded([...recordExpanded, collapseRows[rowId].rule.rule_id]);
-    }
     setDisplayedRows(collapseRows);
   };
 
-  const buildFilteredRows = (allRows, filters) =>
-    allRows
+  const buildFilteredRows = (allRows, filters) => {
+    const expandedRows = displayedRows
+      .filter((ruleExpanded) => ruleExpanded?.isOpen === true)
+      .flatMap((object) => [object?.rule?.rule_id]);
+
+    return allRows
       .filter((rule) => passFilters(rule, filters))
       .map((value, key) => [
         {
           rule: value,
-          isOpen:
-            isAllExpanded ||
-            value?.rule_id === recordExpanded.includes(value?.rule_id),
+          isOpen: isAllExpanded || expandedRows?.includes(value?.rule_id),
           cells: [
             {
               title: (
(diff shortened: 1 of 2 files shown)


gkarat:
There is no "enh" commit type. Refer to https://github.com/angular/angular/blob/main/CONTRIBUTING.md#type and https://github.com/semantic-release/semantic-release#commit-message-format.

src/Components/ClusterRules/ClusterRules.spec.ct.js
```diff
@@ -186,9 +186,9 @@ describe('cluster rules table', () => {
   });
 
   describe('defaults', () => {
-    it('only first item expanded', () => {
```
gkarat, line 189: This is expected that the first item is expanded by default, you can't edit the test like this.

src/Components/ClusterRules/ClusterRules.spec.ct.js
```diff
@@ -207,6 +207,16 @@ describe('cluster rules table', () => {
     cy.get(EXPANDABLES).should('have.length', 0);
   });
 
+  it.only('expand one row then sort', () => {
```
gkarat, line 210:
```suggestion
  it('expand one row then sort', () => {
```

src/Components/ClusterRules/ClusterRules.spec.ct.js
```diff
@@ -207,6 +207,16 @@ describe('cluster rules table', () => {
...
+    cy.get(ROOT).find('#expandable-toggle0').click();
+    cy.get(ROOT)
+      .find('th[data-label=Description]')
+      .find('button')
+      .click()
+      .click();
+    cy.get('table').find('tr').find('td[id="expanded-content11"]');
```
gkarat, line 217: The selector points to a component that is rendered *always* (try to play with a table where all rows are collapsed, the selector will still yield you a container) => this assertion doesn't do what you expect (will be always true).

src/Components/ClusterRules/ClusterRules.js
```diff
@@ -125,7 +127,11 @@ const ClusterRules = ({ cluster }) => {
       .map((value, key) => [
         {
           rule: value,
-          isOpen: isAllExpanded,
+          isOpen: isAllExpanded
+            ? true
+            : value?.rule_id === recordExpanded.includes(value?.rule_id)
+            ? true
+            : false,
```
gkarat, line 134: down with ternary overuse 
```suggestion
          isOpen: isAllExpanded || value?.rule_id === recordExpanded.includes(value?.rule_id);
```

gkarat (commented):
This does fixes the bug, however breaks other things, see the video:

https://user-images.githubusercontent.com/31385370/166456462-b1ced222-e4bb-4e10-99fd-771bb6b6520e.mp4

src/Components/ClusterRules/ClusterRules.spec.ct.js
Fewwy: Its not expanded anymore, so why I shouldn't change the test

gkarat: Showing the first recommendation expanded has been a business rule and a feature. There was no request to change it even though I see that RHEL Advisor stopped doing it for some reason. We need to keep the behavior as it has been.

src/Components/ClusterRules/ClusterRules.js
```diff
@@ -119,13 +118,17 @@ const ClusterRules = ({ cluster }) => {
     setDisplayedRows(collapseRows);
   };
 
-  const buildFilteredRows = (allRows, filters) =>
-    allRows
+  const buildFilteredRows = (allRows, filters) => {
+    const expandedRows = displayedRows
+      .filter((ruleExpanded) => ruleExpanded?.isOpen === true)
+      .flatMap((object) => [object?.rule?.rule_id]);
```
gkarat, line 124:
```suggestion
      .filter((ruleExpanded) => ruleExpanded.isOpen)
      .map((object) => object.rule.rule_id)
```

src/Components/ClusterRules/ClusterRules.js
```diff
@@ -119,13 +118,17 @@ const ClusterRules = ({ cluster }) => {
...
+
+    return allRows
       .filter((rule) => passFilters(rule, filters))
       .map((value, key) => [
         {
           rule: value,
-          isOpen: isAllExpanded,
+          isOpen: isAllExpanded || expandedRows?.includes(value?.rule_id),
```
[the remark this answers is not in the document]
Fewwy: Implemented

src/Components/ClusterRules/ClusterRules.js
Fewwy: Implemented

PR #255, opened 2022-05-05
Include filtering condition in filtersConf

ikerreyes:


src/Components/ClusterRules/ClusterRules.spec.ct.js
```diff
@@ -40,45 +40,29 @@ const filtersConf = {
     selectorText: 'Description',
     values: ['Lorem IPSUM', '1Lorem', 'Not existing recommendation'],
     type: 'input',
+    f: (it, value) =>
+      it.description.toLowerCase().includes(value.toLowerCase()),
```
gkarat, line 44: maybe it is worth abstracting `it` from the rule content, and assuming that it already contains a value that we want to compare.

for example, here, and in other places, you would use just `it` instead of looking for particular params like `it.total_risk`, `it.tags` and etc.

src/Components/ClusterRules/ClusterRules.spec.ct.js
```diff
@@ -40,45 +40,29 @@ const filtersConf = {
...
-      );
-    }
-    // if length is already 0, exit
-    if (filteredData.length === 0) {
-      break;
-    }
-  }
-  return filteredData;
-}
+const filterData = _.partial(filter, filtersConf);
```
gkarat, line 65: Maybe arrow functions could look more intuitive in this case (at least for me, who is not that advanced user of lodash :D). 

```suggestion
const filterData = (data, filters) => filter(filtersConf, data, filters);
```

gkarat (commented):
hey, generally, looks OK to me. to be honest, it starts to be harder to navigate within this code. but if you could complement it with some docs or just comments, that would help.

src/Components/ClusterRules/ClusterRules.spec.ct.js
ikerreyes: I do not understand. Different `f` will need to access different parameters as you pointed out. How would you handle that?

src/Components/ClusterRules/ClusterRules.spec.ct.js
ikerreyes: ok, fair point

ikerreyes:
@gkarat with docs or comments, you refer to document the filter function or the `filtersConf` object and its fields?

gkarat:
> @gkarat with docs or comments, you refer to document the filter function or the `filtersConf` object and its fields?

After re-reading the PR, I'd say this code is actually OK and self-explanatory; please ignore my last comment. (However, you could, for example, in `f` methods that `it` is an instance of rule content).

Something not directly related to this PR: I have one idea to document the functions that can be possibly utilized in multiple specs (for example, https://github.com/RedHatInsights/ocp-advisor-frontend/tree/master/cypress/utils, or the `filter` function from this PR) and can be helpful for other developers (imagine someone else contributing to cypress specs and wish to know which instruments he has besides standard cypress API without reading the implementation of the methods). Maybe, JSdoc can serve well here: https://devhints.io/jsdoc, https://jsdoc.app/. One other benefit is that this can later let us generate HTML documentation. They also have typedef https://jsdoc.app/tags-typedef.html for custom types, so we could document complex types like configs, rule content, and so on. I think we could explore this or just keep it in mind for the nearest future...

ikerreyes:
I agree, better documentation is helpful and needed. However, for this case I do not know which is the best way, because the `filtersConf` object has different properties used for different functions. After checking your links, do you think a `Typedef` would be the way to go to document `filtersConf`?

gkarat:
@ikerreyes , yeah, I think it is even the only way to describe such objects there. I didn't use JSdoc that thoroughly yet, but you can experiment with some type annotations for the most frequently used constructions in your code 🙂 And I'll most probably do the same with such constructions (like the rule content) in a source code

PR #257, opened 2022-05-05
Combine URL tests in RecsListTable

ikerreyes:


gkarat (approved):
nicely done!  as the next idea, it may worth reworking the same way for ClustersListTable (it also utilizes filter query params)

PR #261, opened 2022-05-06
fix(Cluster rules table): display table rows only when filtering is completed

gkarat:
The bug was discovered when the cypress rows counting test was failing (for a manual test, the bug is not visible, however, automated tests are able to identify them since they manipulate with interface way faster than users). 

This makes the table show rows only when the data filtering and sorting are complete.


src/Components/Cluster/Cluster.spec.ct.js
```diff
@@ -63,7 +64,7 @@ describe('cluster page', () => {
     // renders table component
     cy.get(RULES_TABLE).should('have.length', 1);
     // test how many rows were rendered
-    // TODO use checkRowCounts (currently not working, probably related to federated modules)
+    checkRowCounts(singleClusterPageReport.report.data.length);
     cy.getAllRows().should(
```
ikerreyes, line 68: If we already use `checkRowCounts` we can get rid of `cy.getAllRows().should(...`

ikerreyes (approved):
Good finding

Fewwy (approved):
LGTM!

PR #262, opened 2022-05-10
Minor improvements to tests

ikerreyes:
- Add expect URL for ClustersListTable default sorting
- Created function to verify table sorting
- Reduce ROOT usage and use TABLE component
- [ClustersListTable] rename props and data variables
- Removed old hack (CCXDEV-7192)
- Use column name instaed of index

### DIFF
--- a/src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
+++ b/src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
@@ -17,6 +17,7 @@ import {
   MODAL,
   CHECKBOX,
   TBODY,
+  TABLE,
 } from '../../../cypress/utils/components';
 import {
   DEFAULT_ROW_COUNT,
@@ -111,7 +112,7 @@ describe('non-empty successful affected clusters table', () => {
   it('renders table', () => {
     cy.get(ROOT).within(() => {
       cy.get(TOOLBAR).should('have.length', 1);
-      cy.get('table').should('have.length', 1);
+      cy.get(TABLE).should('have.length', 1);
       cy.ouiaType('RHI/TableToolbar').should('have.length', 1);
     });
   });
@@ -182,7 +183,7 @@ describe('non-empty successful affected clusters table', () => {
         `${filterData().length} selected`
       );
       // checks all rows
-      cy.get('table')
+      cy.get(TABLE)
         .find(TBODY)
         .find(ROW)
         .each((row) => {
@@ -206,7 +207,7 @@ describe('non-empty successful affected clusters table', () => {
     it('checkbox is unselected when a row is unselected', () => {
       cy.ouiaId(BULK_SELECT).find('input').click().should('be.checked');
       // removing one row unselects it
-      cy.get('table')
+      cy.get(TABLE)
         .find(TBODY)
         .find(ROW)
         .first()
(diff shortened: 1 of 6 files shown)


src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -170,9 +169,8 @@ describe('non-empty successful affected clusters table', () => {
     });
 
     it('sorting using last seen', () => {
-      cy.get(ROOT)
-        .find('th[data-key=2]') // TODO use column name with data-label
-        .should('have.class', 'pf-c-table__sort pf-m-selected');
+      const column = 'Last seen';
+      tableIsSortedBy(column);
```
gkarat, line 173: perhaps
```suggestion
      tableIsSortedBy('Last seen');
```
?

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -208,7 +206,7 @@ describe('non-empty successful affected clusters table', () => {
     it('checkbox is unselected when a row is unselected', () => {
       cy.ouiaId(BULK_SELECT).find('input').click().should('be.checked');
       // removing one row unselects it
-      cy.get(ROOT)
+      cy.get('table')
```
gkarat, line 209: Isn't it better to keep it under one variable ROOT?

src/Components/ClustersListTable/ClustersListTable.spec.ct.js
```diff
@@ -160,11 +156,11 @@ describe('clusters list table', () => {
...
 
     it('sorting using last seen', () => {
-      // TODO create a function used also in other tests
-      cy.get(ROOT)
-        .find('th[data-label="Last seen"]')
-        .should('have.class', 'pf-c-table__sort pf-m-selected');
-      // TODO check window.location as in  RecsListTable (if applicable)
+      const column = 'Last seen';
+      tableIsSortedBy(column);
+      expect(window.location.search).to.contain(
```
gkarat, line 161: isn't https://github.com/RedHatInsights/ocp-advisor-frontend/pull/262/files#diff-11510ebd9fcd477bd2c0ab72124a29e1b5a7a8ec9743a472cbc67852ba71b74aR173 supposed to also have this check?

ikerreyes: which check? The `window.location`? It is not implemented for the other component

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
gkarat: My argument was rather about using a constant instead of a value directly so that you don't need to change each occurrence once you decide to replace the `'table'` with another selector. But since it's least likely, I think it's not an issue :)

src/Components/ClustersListTable/ClustersListTable.spec.ct.js
gkarat: yep, I meant window.location check. and yes, it's not implemented there, sorry for a noise :D

gkarat (approved):
looks good!

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
ikerreyes: It is a fair point. I will replace them with a new global

PR #263, opened 2022-05-10
fix(RecsListTable): Fixing the category sorting

Fewwy:
Fixed a bug that caused sorting in the category column to not work.
Wrote a test for it.

### DIFF
--- a/src/Components/RecsListTable/RecsListTable.js
+++ b/src/Components/RecsListTable/RecsListTable.js
@@ -247,7 +247,10 @@ const RecsListTable = ({ query }) => {
         },
       ]);
   };
-
+  /* the category sorting compares only the first element of the array.
+   Could be refactored later when we assign a priority numbers to each of the category
+   and sort them in the array based on the priority.
+*/
   const buildDisplayedRows = (rows, index, direction) => {
     const sortingRows = [...rows].sort((firstItem, secondItem) => {
       const d = direction === SortByDirection.asc ? 1 : -1;
--- a/src/Components/RecsListTable/RecsListTable.js
+++ b/src/Components/RecsListTable/RecsListTable.js
@@ -254,9 +254,13 @@ const RecsListTable = ({ query }) => {
       const fst = firstItem[0].rule[RECS_LIST_COLUMNS_KEYS[index]];
       const snd = secondItem[0].rule[RECS_LIST_COLUMNS_KEYS[index]];
       if (index === 3) {
-        return extractCategories(fst)[0].localeCompare(
+        const x = extractCategories(fst)[0].localeCompare(
           extractCategories(snd)[0]
         );
+        const y = extractCategories(snd)[0].localeCompare(
+          extractCategories(fst)[0]
+        );
+        return x > y ? d : y > x ? -d : 0;
       }
       return fst > snd ? d : snd > fst ? -d : 0;
     });
--- a/src/Components/RecsListTable/RecsListTable.js
+++ b/src/Components/RecsListTable/RecsListTable.js
@@ -254,13 +254,10 @@ const RecsListTable = ({ query }) => {
       const fst = firstItem[0].rule[RECS_LIST_COLUMNS_KEYS[index]];
       const snd = secondItem[0].rule[RECS_LIST_COLUMNS_KEYS[index]];
       if (index === 3) {
-        const x = extractCategories(fst)[0].localeCompare(
+        const categoryCompare = extractCategories(fst)[0].localeCompare(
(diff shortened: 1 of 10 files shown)


src/Components/RecsListTable/RecsListTable.js
```diff
@@ -254,9 +254,10 @@ const RecsListTable = ({ query }) => {
       const fst = firstItem[0].rule[RECS_LIST_COLUMNS_KEYS[index]];
       const snd = secondItem[0].rule[RECS_LIST_COLUMNS_KEYS[index]];
       if (index === 3) {
-        return extractCategories(fst)[0].localeCompare(
+        const categoryCompare = extractCategories(fst)[0].localeCompare(
           extractCategories(snd)[0]
         );
+        return categoryCompare === 1 ? d : categoryCompare === -1 ? -d : 0;
```
gkarat, line 260: WDYT?

```suggestion
        return (
          d * extractCategories(fst)[0].localeCompare(extractCategories(snd)[0])
        );
```

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -424,6 +424,27 @@ describe('successful non-empty recommendations list table', () => {
...
+      cy.sortByCol(2).then(() => {
+        expect(window.location.search).to.contain(
+          'sort=-tags&limit=20&offset=0&impacting=true&rule_status=enabled'
+        );
+      });
+      cy.getAllRows()
+        .eq(0)
+        .find('td[data-label="Category"]')
+        .should('contain', 'Service Availability');
+    });
```
gkarat, line 447: This must be instead added to the tests on the line 372

gkarat:
@ikerreyes , could you please verify the change in the tests?

gkarat (approved):
looks good, if Iker sees no problems here, I am OK with merging it

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -368,10 +368,15 @@ describe('successful non-empty recommendations list table', () => {
...
-      ['description', 'publish_date', 'total_risk', 'impacted_clusters_count'],
-      ['Name', 'Modified', 'Total risk', 'Clusters'] // TODO use TABLE_HEADERS
+      [
+        'description',
+        'publish_date',
+        'tags',
+        'total_risk',
+        'impacted_clusters_count',
+      ],
+      ['Name', 'Modified', 'Category', 'Total risk', 'Clusters'] // TODO use TABLE_HEADERS
```
ikerreyes, line 379: Here you can already use `TABLE_HEADERS`

src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -399,12 +404,17 @@ describe('successful non-empty recommendations list table', () => {
                 )
               );
           }
-
+          const filteringTags = filterData(DEFAULT_FILTERS).map((obj) => ({
+            ...obj,
+            tags: obj.tags.filter((string) =>
+              Object.keys(RULE_CATEGORIES).includes(string)
+            ),
+          }));
```
ikerreyes, line 412: Nice idea, but I see 2 issues:
1- If I am not mistaken, this code will modify the underlying objects in the data, thus it can affect the test afterwards. We need to create the modified data object outside the test, so all other tests can make use of it, or inside the test; but using a copy, so the data used by the remaining tests is unaltered
2- These code does not reflect exactly what the code does. Simply add "performace" to `"rule_id": "external.rules.rule_n_two|ERROR_KEY_N42"` and the test will fail. That is because you are comparing by an array of elements in the test, but in the code you use only the first one. The test should reflect what is the expected behavior.
You need to take the first element of each array to fix issue 2 and create a copy of the object to fix issue 1.

ikerreyes (commented):
I propose some changes to address the issues I mentioned in https://github.com/Fewwy/ocp-advisor-frontend/pull/1
There might still be bugs, I would recommend to expand the data we have for this test. Now or maybe later.

src/Components/RecsListTable/RecsListTable.spec.ct.js
ikerreyes: Thanks @Fewwy for the clarification, I was wrong with my assumption on 1

Fewwy: Thanks for your help :) tests are HAAARD!

ikerreyes (approved):

gkarat (approved):

PR #272, opened 2022-05-16
Render cluster version in the affected clusters table

gkarat:
The PR makes the single recommendation view render the **cluster version** column. The value is contained in the API response (`GET /clusters_detail`) within the `meta.cluster_version` path.

Mocked in https://marvelapp.com/prototype/g70a1he/screen/86034747.
Implements https://issues.redhat.com/browse/CCXDEV-6435.

- [x] Render the cluster version obtained from API
- [x] Add sorting to the new column
- [x] Add the checkbox filter to the version parameter
- [x] Update tests

Screenshot 2022-05-18 at 10-28-41 Workloads are using the deprecated PodSecurityPolicy API - Recommendations - OCP Advisor Red Hat Insights

Screenshot 2022-05-18 at 10-29-52 Workloads are using the deprecated PodSecurityPolicy API - Recommendations - OCP Advisor Red Hat Insights

Screenshot from 2022-05-18 10-31-37

Screenshot from 2022-05-18 10-32-01

### DIFF
--- a/src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
+++ b/src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
@@ -96,6 +96,18 @@ describe('test data', () => {
       expect(arr).to.include('not existing cluster');
     });
   });
+  _.uniq(_.flatten(VERSION_COMBINATIONS)).map((c) =>
+    it(`has at least one cluster with version ${c}`, () => {
+      cy.wrap(_.filter(data, (it) => it.meta.cluster_version === c))
+        .its('length')
+        .should('be.gte', 1);
+    })
+  );
+  it(`has at least one cluster without a version`, () => {
+    cy.wrap(_.filter(data, (it) => it.meta.cluster_version === ''))
+      .its('length')
+      .should('be.gte', 1);
+  });
 });
 
 describe('non-empty successful affected clusters table', () => {
@@ -458,6 +470,7 @@ describe('non-empty successful affected clusters table', () => {
   });
 
   describe('filtering', () => {
+    // TODO: use filtersConf approach
     it('no chips are displayed by default', () => {
       cy.get(CHIP_GROUP).should('not.exist');
       cy.get('button').contains('Reset filters').should('not.exist');
--- a/src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
+++ b/src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
@@ -717,8 +717,8 @@ describe('empty failed affected clusters table', () => {
   });
 
   it('renders table header', () => {
-    cy.get(TABLE).find('th').children().eq(0).should('have.text', 'Name');
-    cy.get(TABLE).find('th').children().eq(1).should('have.text', 'Version');
-    cy.get(TABLE).find('th').children().eq(2).should('have.text', 'Last seen');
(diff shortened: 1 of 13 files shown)


src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -375,60 +386,75 @@ describe('non-empty successful affected clusters table', () => {
...
-              'have.length',
-              Math.min(DEFAULT_ROW_COUNT, data.length)
-            );
-            if (order === 'ascending') {
-              cy.get(header).find('button').click();
-            } else {
-              cy.get(header).find('button').dblclick();
-            }
+    _.zip(
+      ['name', 'cluster_version', 'last_checked_at'],
```
ikerreyes, line 390: Maybe we can use here `meta.cluster_version` as in https://github.com/RedHatInsights/ocp-advisor-frontend/blob/master/src/Components/ClustersListTable/ClustersListTable.spec.ct.js#L228 to avoid having to modify the code of the test itself, except for
```if (it.meta.cluster_version === '') {
              it.meta.cluster_version = '0.0.0';
            }```

File: src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js [hunk 2]
```diff
@@ -647,6 +705,7 @@ describe('empty failed affected clusters table', () => {
 
   it('renders table header', () => {
```
ikerreyes, line 706: I think we should start to use a TABLE_HEADERS constant here like https://github.com/RedHatInsights/ocp-advisor-frontend/blob/master/src/Components/ClustersListTable/ClustersListTable.spec.ct.js#L60

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -463,6 +489,38 @@ describe('non-empty successful affected clusters table', () => {
       });
     });
 
+    VERSION_COMBINATIONS.forEach((vs) => {
```
ikerreyes, line 492: Now that we have 2 tests for filtering, maybe we should move to the pattern in other pages, a filtersConf object and have tests for single filters and combinations of filters.

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
```diff
@@ -42,6 +47,12 @@ import { AFFECTED_CLUSTERS_COLUMNS } from '../../AppConstants';
 const ROOT = 'div[id=affected-list-table]';
 const BULK_SELECT = 'clusters-selector';
 const SEARCH_ITEMS = ['ff', 'CUSTOM', 'Foobar', 'Not existing cluster'];
+const VERSION_COMBINATIONS = [
```
ikerreyes, line 50: This is used in the tests, and some tests rely on certain outputs, so we should make sure that we check for that in the data tests. E.g. if we expect that there is at least one cluster with version `'4.18.12'`

ikerreyes (commented):
Nice changes, dropped some ideas for tests improvements. We can address them separately if you prefer.

src/Components/AffectedClustersTable/AffectedClustersTable.spec.ct.js
gkarat: yes, this makes sense. I'll just drop a TODO to not to overwhelm the PR

Issue #35, opened 2021-09-15
Use @babel/plugin-transform-runtime to transform @patternfly imports


gkarat:
Importing components from the @patternfly-based libraries and their `/dist` paths is quite overhead. We should consider a more elegant way of how to deal with that. 

I have come across this approach https://github.com/RedHatInsights/ansible-platform-dashboard/blob/6113d16d025fcc51fc607d6d290317ec27206f08/babel.config.js#L23. Personally, it looks nice and legit. It would be nice to adopt it in the ocp-advisor or come up with an even better idea.

PR #279, opened 2022-05-16
fix(App): babel transformation for the react-core modules

Fewwy:
Updated the react-core imports
Size became smaller (13.02 -> 12.98)
I think that we need to reduce the bundle size overall, it's too big :(

image


gkarat (commented):
hey I've just skimmed through the remaining imports: looks like you haven't included all of the component imports. I am also curious if you can transform imports for `layout`?

image

babel.config.js
```diff
@@ -1,3 +1,13 @@
+const path = require('path');
+const glob = require('glob');
+
+const mapper = {
```
gkarat, line 4: why do we need it?

gkarat: can't we use this mapper to be able to transform imports for `layouts`, `TooltipPosition` and other stuff that has different path?

babel.config.js
```diff
@@ -33,7 +26,7 @@ module.exports = {
               path.resolve(
                 __dirname,
                 `./node_modules/@patternfly/react-core/dist/js/**/${
-                  mapper[importName] || importName
+                  importName || importName
```
gkarat, line 29: what's the meaning of OR operator here?

src/Components/Recommendation/Recommendation.js
```diff
@@ -25,12 +25,12 @@ import {
   Dropdown,
   DropdownItem,
   DropdownToggle,
+  Flex,
+  FlexItem,
 } from '@patternfly/react-core/';
```
gkarat, line 30:
```suggestion
} from '@patternfly/react-core';
```

src/Components/AffectedClustersTable/AffectedClustersTable.js
```diff
@@ -9,16 +9,17 @@ import PrimaryToolbar from '@redhat-cloud-services/frontend-components/PrimaryTo
...
-import { Tooltip } from '@patternfly/react-core/dist/js/components/Tooltip';
 import {
+  Card,
+  CardBody,
+  Tooltip,
+  Bullseye,
   Pagination,
   PaginationVariant,
-} from '@patternfly/react-core/dist/js/components/Pagination/Pagination';
+} from '@patternfly/react-core/';
```
gkarat, line 19:
```suggestion
} from '@patternfly/react-core';
```

gkarat: hey this is not the only place I found trailing slash.. please check all where it occurs and clean it please. otherwise LGTM

gkarat:
Fixes #35

PR #297, opened 2022-05-30
Deploy to production (prod-stable)

gkarat:


gkarat:
Needs confirmation from QE first on the versions feature

ikerreyes (approved):
Seems that all clusters are compliant with SemVer

PR #298, opened 2022-05-31
Avoid hardcoded values

ikerreyes:
Improve different test to avoid hardcoded values in them

### DIFF
--- a/src/Components/RecsListTable/RecsListTable.spec.ct.js
+++ b/src/Components/RecsListTable/RecsListTable.spec.ct.js
@@ -334,7 +334,7 @@ describe('successful non-empty recommendations list table', () => {
       .and('have.text', 'Enabled');
   });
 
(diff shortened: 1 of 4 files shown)


src/Components/RecsListTable/RecsListTable.spec.ct.js
```diff
@@ -221,6 +221,10 @@ describe('data', () => {
       filterData(DEFAULT_FILTERS).length
     );
   });
+  it('there is at least one disabled recommendation in the first page', () => {
+    const firstData = filterData({}).slice(0, DEFAULT_ROW_COUNT + 1);
```
gkarat, line 225: why `+1`? wouldn't it include one extra row from page 2?

ikerreyes: Yes, you are right :man_facepalming:

PR #300, opened 2022-06-02
refactor(ReportDetails): replaced the component with the library one

Fewwy:
Replaced ReportDetails component with the one provided by the library


gkarat (commented):
Missing template definitions, please. here is how template processing works https://github.com/RedHatInsights/frontend-components/blob/master/packages/advisor-components/src/TemplateProcessor/TemplateProcessor.tsx

gkarat (changes requested):
thanks, this looks decent. ~I was thinking about the last piece probably missing here: a rule content can (but not must) have a `node_id` in its content which contains an identifier for the corresponding knowledgebase article. We need to pass it (if available) to `ReportDetails` too via `kbaDetails` prop, so that the report contains a "Knowledgebase Article" section.~

gkarat:
Could you please also remove the Federated modules title from https://github.com/RedHatInsights/ocp-advisor-frontend/blob/master/ARCHITECTURE.md#federated-modules?

gkarat (approved):

PR #301, opened 2022-06-02
[CCXDEV-8214] Homogenize the use of data

ikerreyes:
- data is origional data, sorted with default and enhanced as needed
- dataUnsorted is the same as data but keeping the original ordering (useful for sorting test)
- filterData functions follow the same signature

### DIFF
--- a/src/Components/ClusterRules/ClusterRules.spec.ct.js
+++ b/src/Components/ClusterRules/ClusterRules.spec.ct.js
@@ -64,7 +64,7 @@ const filtersConf = {
   },
 };
 
-const filterData = (data, filters) => filter(filtersConf, data, filters);
+const filterData = (filters) => filter(filtersConf, data, filters);
 const filterApply = (filters) => applyFilters(filters, filtersConf);
 
 // TODO add more combinations of filters for testing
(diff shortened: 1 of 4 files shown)


src/Components/ClusterRules/ClusterRules.spec.ct.js
```diff
@@ -64,7 +64,7 @@ const filtersConf = {
   },
 };
 
-const filterData = (data, filters) => filter(filtersConf, data, filters);
+const filterData = (filters) => filter(filtersConf, data, filters);
```
gkarat, line 67: I think here you pass `data` by reference, not a value (objects and arrays in JS passed by reference in functions). It didn't lead to problems in tests. However, it may create issues in the future...

src/Components/ClustersListTable/ClustersListTable.spec.ct.js
```diff
@@ -118,62 +119,54 @@ const filtersConf = {
...
-      _.filter(filterData(data), (it) => it.last_checked_at === undefined)
+      _.filter(filterData(), (it) => it.last_checked_at === undefined)
     ).to.have.length.gte(1);
   });
-  it('at least one entry does not have all values for total risk categories', () => {
+  it('at least one entry in the original data does not have all values for total risk categories', () => {
     expect(
       _.filter(
-        filterData(clusters['data']),
+        filterData(DEFAULT_FILTERS, clusters['data']),
```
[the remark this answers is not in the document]
gkarat: you are true, there will be a problem passing the 2nd argument. ignore me then :)

src/Components/ClusterRules/ClusterRules.spec.ct.js
ikerreyes: That's why I didn't put too much attention to this. However, it might help to avoid some troubles, so I will change it.

ikerreyes: @gkarat Do you think it is better now to have `values` as `let` and `data` as `const`? Is it more clear?

PR #309, opened 2022-06-06
Fix wrong placement of "Expand all" button in tables

gkarat:
Fixes https://issues.redhat.com/browse/CCXDEV-8191.

Moves the "arrow" button for expanding/hovering all rows from the toolbar to table headers.

Screenshot 2022-06-06 at 16-10-17 Cluster With Issues - Clusters - OCP Advisor Red Hat Insights

Screenshot 2022-06-06 at 16-10-24 Recommendations - OCP Advisor Red Hat Insights

### DIFF
--- a/cypress/utils/table.js
+++ b/cypress/utils/table.js
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 
-import { ROW, TBODY } from './components';
+import { ROW, TABLE_HEADER, TBODY } from './components';
 
 function checkTableHeaders(expectedHeaders) {
   /* patternfly/react-table-4.71.16, for some reason, renders extra empty `th` container;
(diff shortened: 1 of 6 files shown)


gkarat:
@RedHatInsights/team-interact, @ikerreyes, PTAL 🐊

cypress/utils/table.js
```diff
@@ -33,9 +33,14 @@ function tableIsSortedBy(columnTitle) {
     .should('have.class', 'pf-c-table__sort pf-m-selected');
 }
 
+function expandAllRows() {
```
ikerreyes, line 36: Naming can be a bit confusing. If the button is already clicked, it will collapse the rows instead of expanding them. What about `expandCollapseAllRows()`? Or maybe even better, we define a component `rowToggler = `${TABLE_HEADER} .pf-c-table__toggle` and then instead of a function, we simple use it in the tests as `cy.get(rowToggler).click()`?

ikerreyes (approved):

cypress/utils/table.js
gkarat: @ikerreyes, yes, why not, I think it looks better

PR #311, opened 2022-06-07
[CCXDEV-7133] Bulk selector tests improve

ikerreyes:
- Check disabling multiple clusters button is also modified
- Test for test of selector


gkarat (approved):
looks OK, thank you

PR #325, opened 2022-06-14
refactor(Empty states): set proper OUIA attributes to empty state containers

gkarat:
Fixes https://issues.redhat.com/browse/CCXDEV-8329.


ikerreyes (approved):
Thanks!

PR #328, opened 2022-06-14
Share code for filtering tests

ikerreyes:
- Create a function for filtering tests that wraps shared code
- Use function in single filter tests
- Use function in combined filter tests

[CCXDEV-8215](https://issues.redhat.com/browse/CCXDEV-8215)


gkarat (approved):
great work done! I liked the way we managed to make these generic functions. Given the new discussions of standardizing tables UX across c.r.c., I think your efforts can be applied in the broader context soon :)

PR #346, opened 2022-07-11
fix(OCP Advisor): changed N/A to not available

Fewwy:


gkarat (changes requested):
hey, looks OK, but we have a redundant https://github.com/RedHatInsights/ocp-advisor-frontend/blob/4f809dacc113a1d89a70e44b56a7305ae186205a/src/Messages.js#L608: can we leave only one message descriptor that we will use across the app? so that have either only 'nA' message or only 'notAvailable'? there should be no places in app where 'N/A' will appear

PR #347, opened 2022-07-11
refactor: remove trailing slashes in endpoint paths

gkarat:
This removes trailing slashes from endpoints. Before, it was causing 301 responses.


Fewwy (approved):
@gkarat LGTM!

PR #359, opened 2022-07-14
refactor(Recommendations list): Map sorting search parameters from column names

gkarat:
Fixes https://issues.redhat.com/browse/CCXDEV-8064.

This changes the way how Advisor builds sorting parameters in the search URL query. The name for each parameter is taken from the column name instead of using the data field name from the API response. The PR also refactors the sorting function for the recommendations list table: the algorithm is now the same as for the clusters list table, and a switch-case clause is used.


ikerreyes (approved):
Much better, now both tests are consistent and are quite simple. And the columnName2UrlParam is more meaningful now :)

Fewwy (approved):
LGTM!

PR #366, opened 2022-07-20
feat(ClusterRules): Add impacted to single Cluster page

Fewwy:
Added the impacted column component + state when the impacted data is not available + tests

image

image

### DIFF
--- a/src/Components/ClusterRules/ClusterRules.js
+++ b/src/Components/ClusterRules/ClusterRules.js
@@ -178,17 +178,36 @@ const ClusterRules = ({ cluster }) => {
                 </div>
               ),
             },
-            {
-              title: (
-                <div key={key}>
-                  <DateFormat
-                    date={value.impacted}
-                    type="relative"
-                    tooltipProps={{ position: TooltipPosition.bottom }}
-                  />
-                </div>
-              ),
-            },
+            value.impacted
+              ? {
+                  title: (
+                    <div key={key}>
+                      <DateFormat
+                        extraTitle={`${intl.formatMessage(
+                          messages.impacted
+                        )}: `}
+                        date={value.impacted}
+                        type="relative"
+                        tooltipProps={{ position: TooltipPosition.bottom }}
+                      />
+                    </div>
+                  ),
+                }
+              : {
+                  title: (
+                    <Tooltip
+                      key={key}
+                      content={
+                        <span>
+                          {intl.formatMessage(messages.impacted) + ': '}
+                          {intl.formatMessage(messages.nA)}
+                        </span>
+                      }
+                    >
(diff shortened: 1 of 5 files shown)


ikerreyes (approved):
Looks good to me. 

Just one note, the field is declared as optional in the API spec (https://github.com/RedHatInsights/insights-results-smart-proxy/blob/master/server/api/v2/openapi.json#L1229). I think it is an error, as it does not make sense to have a rule hit without a timestamp, but if it is possible, we should enhance the test to account for that.

I will try to clarify that.

src/Components/ClusterRules/ClusterRules.js
```diff
@@ -244,6 +274,12 @@ const ClusterRules = ({ cluster }) => {
       sortingRows = [...rows].sort((firstItem, secondItem) => {
         const fst = firstItem[0].rule[CLUSTER_RULES_COLUMNS_KEYS[index]];
         const snd = secondItem[0].rule[CLUSTER_RULES_COLUMNS_KEYS[index]];
+        if (index === 3) {
```
gkarat, line 277: please don't use hard-coded values here. Could you please map column indices with the analogy to https://github.com/RedHatInsights/ocp-advisor-frontend/blob/acba19655f0b803ee934a2c226b1489170dd2ca5/src/AppConstants.js#L264 ?

then you could use it here in a way:
```suggestion
        if (index === CLUSTER_RULES_IMPACTED_CELL) {
```

src/Components/ClusterRules/ClusterRules.js
```diff
@@ -244,6 +274,12 @@ const ClusterRules = ({ cluster }) => {
...
+          //sorting for the impacted column
+          const firstDate = new Date(firstItem[0].rule.impacted || 0);
+          const secondDate = new Date(secondItem[0].rule.impacted || 0);
+          return firstDate > secondDate ? d : secondDate > firstDate ? -d : 0;
```
gkarat, line 281: This is the same as for the line two lines below. What about declaring `fst` and `snd` with `let`, and use if-else clause with a condition `if (index === CLUSTER_RULES_IMPACTED_CELL)`, where you will then decide which values to assign to `fst` and `snd`?..

src/Components/ClusterRules/ClusterRules.js
Fewwy: Great suggestion. Thanks

src/Components/ClusterRules/ClusterRules.js
Fewwy: I applied your suggestion. Thanks

gkarat (approved):

PR #372, opened 2022-07-25
feat(ClusterRules): Improved test for the Cluster Rules

Fewwy:


src/Components/ClusterRules/ClusterRules.cy.js
```diff
@@ -122,6 +122,9 @@ describe('test data', () => {
     expect(filteredData).to.have.length.gte(1);
     expect(filteredData).to.have.length.lt(RULES_ENABLED);
   });
+  it('has 1 rule without impacted field', () => {
```
ikerreyes, line 125: I feel this too much restricting. I would test for something like, `'at least 1 rule without impacted field'`. The reason is that a new entry without impacted field will make this test to fail, but the tests below do not require one and only one, but any number of rules without impacted should work.

src/Components/ClusterRules/ClusterRules.cy.js
```diff
@@ -211,9 +214,24 @@ describe('cluster rules table', () => {
     ).forEach(([category, label]) => {
       SORTING_ORDERS.forEach((order) => {
         it(`${order} by ${label}`, () => {
+          let sortingParameter = category;
+          // modify sortingParameters for certain values
+          if (category === 'description') {
```
ikerreyes, line 219: If tests where passing previously, and now we are only modifying `impacted`, why do we need to modify the sorting parameter for all the fields and not only for `impacted`?

src/Components/ClusterRules/ClusterRules.cy.js
```diff
@@ -211,9 +214,24 @@ describe('cluster rules table', () => {
...
+            // name sorting is case insensitive
+            sortingParameter = (it) => it.description.toLowerCase();
+          } else if (category === 'created_at') {
+            sortingParameter = (it) =>
+              it.created_at || '1970-01-01T01:00:00.001Z';
+          } else if (category === 'impacted') {
+            sortingParameter = (it) =>
+              it.impacted || '1970-01-01T01:00:00.001Z';
+          } else if (category === 'total_risk') {
+            sortingParameter = (it) => it.total_risk || 3;
```
ikerreyes, line 229: Why default total risk is 3?

src/Components/ClusterRules/ClusterRules.cy.js
```diff
@@ -122,6 +122,9 @@ describe('test data', () => {
     expect(filteredData).to.have.length.gte(1);
     expect(filteredData).to.have.length.lt(RULES_ENABLED);
   });
+  it('has 1 rule without impacted field', () => {
+    expect(_.uniq(_.map(data, 'Not available'))).to.have.length.lengthOf(1);
```
ikerreyes, line 126: I do not see any reference to `impacted` here. What is this code suppose to be doing exactly?

src/Components/ClusterRules/ClusterRules.cy.js
Fewwy: You are correct, removed the "fluff" and left the "if" for the impacted category only

ikerreyes (approved):

gkarat (approved):
looks legit, thank you

gkarat:
And I guess `test` is a more appropriate keyword for this PR than `feat`. Just a note

PR #382, opened 2022-08-04
feat(ClustersList): new tests for url params

Fewwy:
Added new tests for the ClusterList table

I had to add 2 checks to separate how urlParamConvert and hasChip work.
I think that this could be improved

### DIFF
--- a/cypress/utils/filters.js
+++ b/cypress/utils/filters.js
@@ -4,11 +4,6 @@ Utilities related to URL parameters passed for table filtering
 
 import _ from 'lodash';
 
-import {
-  FILTER_CATEGORIES,
-  CLUSTER_FILTER_CATEGORIES,
-} from '../../src/AppConstants';
-
 import { CHIP_GROUP, CHIP } from './components';
 
 const FILTERS_DROPDOWN = 'ul[class=pf-c-dropdown__menu]';
@@ -78,36 +73,16 @@ function applyFilters(filters, filtersConf) {
   }
 }
 
-function urlParamConvert(key, value, type) {
-  if (type === 'recommendation') {
-    const filterCategory = _.find(
-      _.values(FILTER_CATEGORIES),
-      (it) => it.urlParam === key
-    );
-    const title = _.capitalize(filterCategory.title);
-    const label = _.find(filterCategory.values, (it) => it.value === value)
-      .label.props.children;
-    return [title, label];
-  } else if (type === 'cluster') {
-    const filterCategory = _.find(
-      _.values(CLUSTER_FILTER_CATEGORIES),
-      (it) => it.urlParam === 'hits'
-    );
-    const title = _.capitalize(filterCategory.title);
-    const label = _.find(filterCategory.values, (it) => it.value === value)
-      ?.label.props.children;
-    console.log(title);
-    console.log(label);
-    return [title, label];
-  }
+function urlParamConvert(key, value, filters) {
+  const filterCategory = _.find(_.values(filters), (it) => it.urlParam === key);
+  const title = _.capitalize(filterCategory.title);
+  const label = _.find(filterCategory.values, (it) => it.value === value)?.label
+    .props.children;
+  return [title, label];
 }
 
-function hasChip(name, value, type) {
-  if (type === 'recommendation') {
-    cy.contains(CHIP_GROUP, name).parent().contains(CHIP, value);
-  } else if (type === 'cluster') {
-    cy.contains(CHIP_GROUP, name).parent().contains(CHIP, value);
-  }
+function hasChip(name, value) {
+  cy.contains(CHIP_GROUP, name).parent().contains(CHIP, value);
 }
 /**
  * filter data given a set of filter values and their configuration
--- a/cypress/utils/filters.js
+++ b/cypress/utils/filters.js
@@ -4,7 +4,10 @@ Utilities related to URL parameters passed for table filtering
 
 import _ from 'lodash';
 
-import { FILTER_CATEGORIES } from '../../src/AppConstants';
+import {
+  FILTER_CATEGORIES,
+  CLUSTER_FILTER_CATEGORIES,
+} from '../../src/AppConstants';
 
 import { CHIP_GROUP, CHIP } from './components';
 
@@ -75,21 +78,37 @@ function applyFilters(filters, filtersConf) {
   }
 }
 
-function urlParamConvert(key, value) {
-  const filterCategory = _.find(
-    _.values(FILTER_CATEGORIES),
-    (it) => it.urlParam === key
-  );
-  const title = _.capitalize(filterCategory.title);
-  const label = _.find(filterCategory.values, (it) => it.value === value).label
-    .props.children;
-  return [title, label];
+function urlParamConvert(key, value, type) {
+  if (type === 'recommendation') {
+    const filterCategory = _.find(
+      _.values(FILTER_CATEGORIES),
+      (it) => it.urlParam === key
+    );
+    const title = _.capitalize(filterCategory.title);
+    const label = _.find(filterCategory.values, (it) => it.value === value)
+      .label.props.children;
+    return [title, label];
+  } else if (type === 'cluster') {
+    const filterCategory = _.find(
+      _.values(CLUSTER_FILTER_CATEGORIES),
+      (it) => it.urlParam === 'hits'
+    );
+    const title = _.capitalize(filterCategory.title);
+    const label = _.find(filterCategory.values, (it) => it.value === value)
+      ?.label.props.children;
+    console.log(title);
+    console.log(label);
+    return [title, label];
+  }
 }
 
-function hasChip(name, value) {
-  cy.contains(CHIP_GROUP, name).parent().contains(CHIP, value);
+function hasChip(name, value, type) {
+  if (type === 'recommendation') {
+    cy.contains(CHIP_GROUP, name).parent().contains(CHIP, value);
+  } else if (type === 'cluster') {
+    cy.contains(CHIP_GROUP, name).parent().contains(CHIP, value);
+  }
 }
-
 /**
  * filter data given a set of filter values and their configuration
  * @param {@filtersConf} conf - Configuration of the filters
--- a/cypress/utils/filters.js
+++ b/cypress/utils/filters.js
@@ -85,7 +85,7 @@ function urlParamConvert(key, value, filters) {
     label = value;
   } else {
     title = _.capitalize(filterCategory.title);
-    label = _.find(filterCategory.values, (it) => it.value === value)?.label
+    label = _.find(filterCategory.values, (it) => it.value === value).label
       .props.children;
   }
   return [title, label];
--- a/cypress/utils/filters.js
+++ b/cypress/utils/filters.js
@@ -74,10 +74,20 @@ function applyFilters(filters, filtersConf) {
 }
 
 function urlParamConvert(key, value, filters) {
-  const filterCategory = _.find(_.values(filters), (it) => it.urlParam === key);
-  const title = _.capitalize(filterCategory.title);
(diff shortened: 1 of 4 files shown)


cypress/utils/filters.js
```diff
@@ -75,21 +78,37 @@ function applyFilters(filters, filtersConf) {
...
+  } else if (type === 'cluster') {
+    const filterCategory = _.find(
+      _.values(CLUSTER_FILTER_CATEGORIES),
+      (it) => it.urlParam === 'hits'
+    );
+    const title = _.capitalize(filterCategory.title);
+    const label = _.find(filterCategory.values, (it) => it.value === value)
+      ?.label.props.children;
+    console.log(title);
+    console.log(label);
```
adonispuente, line 100: these console.logs made it through

adonispuente (commented):
When trying to run this, I keep getting an error about a tsconfig file not being found. When running im definitely up to date with master. I may be not setting something up properly, i definitely do not have this config file

cypress/utils/filters.js
```diff
@@ -75,21 +78,37 @@ function applyFilters(filters, filtersConf) {
...
+    return [title, label];
+  }
 }
 
-function hasChip(name, value) {
-  cy.contains(CHIP_GROUP, name).parent().contains(CHIP, value);
+function hasChip(name, value, type) {
+  if (type === 'recommendation') {
```
ikerreyes, line 106: what is the difference at each branch of the if?

src/Components/ClustersListTable/ClustersListTable.cy.js
```diff
@@ -199,6 +200,61 @@ describe('data', () => {
...
+    it('recognizes all parameters', () => {
+      const urlSearchParameters = new URLSearchParams(urlParams);
+      for (const [key, value] of urlSearchParameters) {
+        if (key == 'text') {
+          hasChip('Name', value, 'cluster');
+          cy.get('.pf-m-fill > .pf-c-form-control').should('have.value', value);
+        } else {
+          value.split(',').forEach((it) => {
+            const [group, item] = urlParamConvert(key, it, 'cluster');
+            item ? hasChip(group, item, 'cluster') : undefined;
```
ikerreyes, line 245: why is this conditional? Every parameter should be reflected in the UI, otherwise there is an error

cypress/utils/filters.js
```diff
@@ -75,21 +78,37 @@ function applyFilters(filters, filtersConf) {
...
-  const filterCategory = _.find(
-    _.values(FILTER_CATEGORIES),
-    (it) => it.urlParam === key
-  );
-  const title = _.capitalize(filterCategory.title);
-  const label = _.find(filterCategory.values, (it) => it.value === value).label
-    .props.children;
-  return [title, label];
+function urlParamConvert(key, value, type) {
+  if (type === 'recommendation') {
```
ikerreyes, line 82: I do not understand why you need this distinction with an if. What are the differences in both branches? In addition, doing something specific for just 1 tests should be placed inside the test module, not in the generic

ikerreyes (commented):
I do not see the point of the `if` clauses and those changes make the functions no longer generic

src/Components/ClustersListTable/ClustersListTable.cy.js
gkarat: I also don't think the condition is necessary here. the check should pass for all filter/pagination values

cypress/utils/filters.js
```diff
@@ -75,21 +78,37 @@ function applyFilters(filters, filtersConf) {
...
+    const filterCategory = _.find(
+      _.values(FILTER_CATEGORIES),
```
gkarat, line 84: \+ to the Iker's comment

I think, instead of adding an extra prop to this function, you could better focus on `FILTER_CATEGORIES` and `CLUSTER_FILTER_CATEGORIES` (this is the only thing these two branches are different with, right?). Can we get back to the initial version of the function but only replace one line in the following way:

```suggestion
      _.values(FILTER_CATEGORIES.concat(CLUSTER_FILTER_CATEGORIES)),
```

cypress/utils/filters.js
gkarat: burn them!

cypress/utils/filters.js
gkarat: \+, the two branches are identical

gkarat (changes requested):

Fewwy:
> When trying to run this, I keep getting an error about a tsconfig file not being found. When running im definitely up to date with master. I may be not setting something up properly, i definitely do not have this config file

That's very strange we don't have the typescript install in OCP Advisor

cypress/utils/filters.js
ikerreyes: Maybe we can create an even more generic function, so that we pass the `FILTER_CATEGORIES` (in whichever form is required, i.e. `FILTER_CATEGORIES` or `CLUSTER_FILTER_CATEGORIES`). So the function will be something like:

```
function urlParamConvert(key, value, map) {
  const filterCategory = _.find(
    _.values(map),
    (it) => it.urlParam === key
  );
  const title = _.capitalize(filterCategory.title);
  const label = _.find(filterCategory.values, (it) => it.value === value).label
    .props.children;
  return [title, label];
}
```

and then we can create "partials" (sorry for the Python naming) for the different tests

```
urlParamConverter = (key, value) => urlParamConvert(key, value, FILTER_CATEGORIES)
```

gkarat: @ikerreyes, this is absolutely legit, from my point of view. Mine option is "the least effort," whereas yours is somewhat "generic/proper solution striving."

src/Components/ClustersListTable/ClustersListTable.cy.js
```diff
@@ -199,6 +201,65 @@ describe('data', () => {
   });
 });
 
+const urlParamsList = [
```
ikerreyes, line 204: The test data lacks some cases in which hits are a list of values. E.g. `hits=1,2,3`

src/Components/ClustersListTable/ClustersListTable.cy.js
```diff
@@ -199,6 +201,65 @@ describe('data', () => {
...
+    it('recognizes all parameters', () => {
+      const urlSearchParameters = new URLSearchParams(urlParams);
+      for (const [key, value] of urlSearchParameters) {
+        if (key == 'text') {
+          hasChip('Name', value, 'cluster');
+          cy.get('.pf-m-fill > .pf-c-form-control').should('have.value', value);
+        } else {
+          value.split(',').forEach((it) => {
+            const [group, item] = urlParamConvert(
+              'hits',
```
ikerreyes, line 246: Definitely, this is not something you want to do. Basically you are setting all items in the else clause to check for `hits`. That shouldn't be the case.

src/Components/ClustersListTable/ClustersListTable.cy.js
```diff
@@ -199,6 +201,65 @@ describe('data', () => {
...
+              it,
+              CLUSTER_FILTER_CATEGORIES
+            );
+            item ? hasChip(group, item) : undefined;
```
ikerreyes, line 250: This is not a good idea, we want to test that each parameter we want to test should be there. If it is not there, then the is something wrong

src/Components/ClustersListTable/ClustersListTable.cy.js
```diff
@@ -258,7 +319,7 @@ describe('clusters list table', () => {
     });
 
     it('applies total risk "All clusters" filter', () => {
-      hasChip('Total risk', 'All clusters');
+      hasChip('Total risk', 'All clusters', 'cluster');
```
ikerreyes, line 322: Has chips should be agnostic of the page

ikerreyes (changes requested):
In my opinion, you are trying to make tests pass, but we should analyze the root cause for the failures when test are not passing as we expect.

E.g.
```
item ? hasChip(group, item) : undefined;
```
this goes against the nature of the test, as we want to ensure that each item we pass in the URL is reflected in the chips, and this test can prevent us. Maybe in the future with new parameters test just pass, even if something is wrong, because our assertions are not strong enough.

You are totally right that the current `urlParamConvert` function implementation is not correct for the use cases and it needs updated. As you are doing, passing it the configuration for each page (e.g. `urlParamConvert(key, it, FILTER_CATEGORIES);`) is, at least for now, a good solution.

I know tests are still failing, and we should go to the root cause. In this case, it is because the version parameter is not in the configuration. Probably the reason is that it is a very dynamic field that is constructed on the fly. However, if probably has a very easy to interpret mapping: the same field is the chip name and the values are directly mapped. For that, I propose to update the `urlParamConvert` function to something that accepts that schema as well.
What do you think about something like

```
function urlParamConvert(key, value, filters) {
  const filterCategory = _.find(_.values(filters), (it) => it.urlParam === key);
  let title, label;
  if (filterCategory === undefined) {
    title = _.capitalize(key);
    label = value;
  } else {
    title = _.capitalize(filterCategory.title);
    label = _.find(filterCategory.values, (it) => it.value === value).label
      .props.children;
  }
  return [title, label];
}
```
? We just separate the cases when the filter knows about the param passed and when not.

ikerreyes:
I cannot use `hits=all` in the ClustersListTable`. Cypress fails to render, is this expected?

src/Components/ClustersListTable/ClustersListTable.cy.js
Fewwy: yeah, that was a leftover from the first version. Good catch

src/Components/ClustersListTable/ClustersListTable.cy.js
Fewwy: Yeah, that was my way to workaround the issue.

src/Components/ClustersListTable/ClustersListTable.cy.js
Fewwy: I added an array 
`const clusterListFilterValues = ['hits'];`
to store values and utilized _.find() lodash method to search the values in the array

src/Components/ClustersListTable/ClustersListTable.cy.js
ikerreyes: @Fewwy you still have this code of `hasChip` with 3 parameters in many other places

Fewwy: Removed them everywhere

cypress/utils/filters.js
```diff
@@ -75,21 +73,27 @@ function applyFilters(filters, filtersConf) {
   }
 }
 
-function urlParamConvert(key, value) {
+function urlParamConvert(key, value, filters) {
   const filterCategory = _.find(
-    _.values(FILTER_CATEGORIES),
-    (it) => it.urlParam === key
+    _.values(filters),
+    (it) => it?.urlParam === key
```
ikerreyes, line 79: I think you can remove the `?` here as well

src/Components/RecsListTable/Downloadstuff.js
```diff
@@ -0,0 +1,15 @@
+import { useGetRecsQuery } from '../../Services/SmartProxy';
```
ikerreyes, line 1: why is this file needed? What is its purpose?

src/Components/ClustersListTable/ClustersListTable.cy.js
Fewwy: I made a separate Jira card covering this issue

ikerreyes (approved):

Fewwy:
> 

The issue is that the CLUSTER_FILTER_CATEGORIES all categories has a different structure. I think that should be covered in the other task

gkarat (approved):
vincent-price-looks-good-to-me

PR #383, opened 2022-08-05
Fix the "OUIA data safe set to false for cluster without archives"

gkarat:
Fixes https://issues.redhat.com/browse/OCPADVISOR-47.

All tables were incorrectly setting the `data-ouia-safe` attribute to false when there is an error requesting API. Now, they should always be setting it to `true` in all scenarios when no animation and visual change is ongoing (as expected).


ikerreyes (approved):
Nice, thank you very much

PR #389, opened 2022-08-15
fix(Tables): reduce re-renders; fix the incorrect data-ouia-safe attribute

gkarat:
This continues on https://issues.redhat.com/browse/OCPADVISOR-47: fixes the incorrect data-ouia-safe while tables are in the error empty state. Also, the PR rearranges `useEffect`s so that extra re-renders are avoided.


Fewwy (approved):
Seems legit! :D LGTM

PR #402, opened 2022-08-24
Test for justification message when rule is acked

ikerreyes:
Test for checking that justification messages are correctly displayed (if they should)

Fixes [CCXDEV-9144](https://issues.redhat.com/browse/CCXDEV-9144)

### DIFF
--- a/src/Components/Recommendation/Recommendation.cy.js
+++ b/src/Components/Recommendation/Recommendation.cy.js
@@ -284,6 +284,74 @@ describe('recommendation page for disabled recommendation', () => {
   });
 });
 
+describe.only('justification message', () => {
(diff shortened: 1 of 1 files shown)


src/Components/Recommendation/Recommendation.cy.js
```diff
@@ -284,6 +284,74 @@ describe('recommendation page for disabled recommendation', () => {
   });
 });
 
+describe.only('justification message', () => {
```
gkarat, line 287:
```suggestion
describe('justification message', () => {
```

gkarat (approved):
Please, remove `only` and let other tests run too :D Otherwise, looks clean 👍🏼

PR #427, opened 2022-10-05
Ephemeral environments: add build and PR check scripts

gkarat:
Implements https://issues.redhat.com/browse/OCPADVISOR-25 and follows https://consoledot.pages.redhat.com/docs/dev/getting-started/frontend-migration.html#_step_3_add_build_and_pr_check_scripts.

### DIFF
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# --------------------------------------------
+# Export vars for helper scripts to use
+# --------------------------------------------
+# name of app-sre "application" folder this component lives in; needs to match for quay
+export COMPONENT="ocp-advisor"
+# IMAGE should match the quay repo set by app.yaml in app-interface
+export IMAGE="quay.io/cloudservices/ocp-advisor-frontend"
+export WORKSPACE=${WORKSPACE:-$APP_ROOT} # if running in jenkins, use the build's workspace
+export APP_ROOT=$(pwd)
+#16 is the default Node version. Change this to override it.
+export NODE_BUILD_VERSION=16
+COMMON_BUILDER=https://raw.githubusercontent.com/RedHatInsights/insights-frontend-builder-common/master
+
+set -exv
+# source is preferred to | bash -s in this case to avoid a subshell
+source <(curl -sSL $COMMON_BUILDER/src/frontend-build.sh)
+BUILD_RESULTS=$?
+
+# Stubbed out for now
+mkdir -p $WORKSPACE/artifacts
+cat << EOF > $WORKSPACE/artifacts/junit-dummy.xml
+<testsuite tests="1">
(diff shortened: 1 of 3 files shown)


build_deploy.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# --------------------------------------------
+# Export vars for helper scripts to use
+# --------------------------------------------
+# name of app-sre "application" folder this component lives in; needs to match for quay
+export COMPONENT="ocp-advisor"
+# IMAGE should match the quay repo set by app.yaml in app-interface
+export IMAGE="quay.io/cloudservices/ocp-advisor-frontend"
+export WORKSPACE=${WORKSPACE:-$APP_ROOT} # if running in jenkins, use the build's workspace
+export APP_ROOT=$(pwd)
```
ikerreyes, line 11: I now it is what is on the docs but sound a bit weird to define a variable after using it as fallback for `WORKSPACE`. Can we double check with other apps?

pr_check.sh
```diff
@@ -0,0 +1,37 @@
...
+export WORKSPACE=${WORKSPACE:-$APP_ROOT} # if running in jenkins, use the build's workspace
+export APP_ROOT=$(pwd)
+# 16 is the default Node version. Change this to override it.
+export NODE_BUILD_VERSION=16
+COMMON_BUILDER=https://raw.githubusercontent.com/RedHatInsights/insights-frontend-builder-common/master
+
+# --------------------------------------------
+# Options that must be configured by app owner
+# --------------------------------------------
+IQE_PLUGINS="ocp_advisor"
```
ikerreyes, line 19: This will be `ccx`.

pr_check.sh
```diff
@@ -0,0 +1,37 @@
...
+IQE_MARKER_EXPRESSION="smoke"
```
ikerreyes, line 20: This will be `ui` or `ui and smoke` or something. We will need to decide. But it can be done later on.

ikerreyes (approved):

mkholjuraev (approved):
LGTM. All of these are correct according to the docs.

build_deploy.sh
gkarat: this is indeed strange. in other apps we have the same order: https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/build_deploy.sh, https://github.com/RedHatInsights/vulnerability-ui/blob/master/build_deploy.sh for instance

ikerreyes: Yes, well, maybe it is not used if `WORKSPACE` is always defined...

pr_check.sh
gkarat: ok, I left just `ui`, is it ok for now?

ikerreyes: sure, nothing will be run, so it really doesn't matter (yet)

PR #441, opened 2022-10-20
feat(GitHubActions): Add cypress yml config to run github actions

Fewwy:
Tested it on my fork and it works :)
image

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -7,8 +7,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      # Install NPM dependencies, cache them correctly
-      # and run all Cypress tests
+      - name: Install dependencies
+        run: npm install
       - name: Cypress run
         uses: cypress-io/github-action@v4
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/cypress.yml
```diff
@@ -0,0 +1,16 @@
+name: cypress-test-run
+on: [push]
```
ikerreyes, line 2: To me this is good, but if we reach some limits or anything, we can consider alternatives.

ikerreyes (approved):
Great, thanks!

.github/workflows/cypress.yml
Fewwy: Yeah, we can check the https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows 
and choose what suits us better

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+
+  cypress-run:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
gkarat, line 9: It seems that v3 is available (however, not sure what is the difference): https://github.com/actions/checkout

gkarat (approved):
looks good to me, let's merge and see how it works

PR #450, opened 2022-11-10
[OCPADVISOR-6] Reset filters resets only filters

johnsonm325:
When you click on "Reset filters" values set for pagination and sorting get also reset. This is on every table in the app.

This PR ensures that on "Reset filters" click, only filters get reset and not pagination or sorting.

### DIFF
--- a/src/Components/AffectedClustersTable/AffectedClustersTable.cy.js
+++ b/src/Components/AffectedClustersTable/AffectedClustersTable.cy.js
@@ -36,8 +36,10 @@ import {
 import {
   itemsPerPage,
   checkPaginationTotal,
+  checkPaginationSelected,
   checkPaginationValues,
   changePagination,
+  checkCurrentPage,
 } from '../../../cypress/utils/pagination';
 import rule from '../../../cypress/fixtures/api/insights-results-aggregator/v2/rule/external.rules.rule|ERROR_KEY.json';
 import { AFFECTED_CLUSTERS_COLUMNS } from '../../AppConstants';
@@ -520,6 +522,26 @@ describe('non-empty successful affected clusters table', () => {
       cy.get(TOOLBAR).find('button').contains('Reset filters').click();
       cy.get(TOOLBAR).find(CHIP_GROUP).should('not.exist');
       checkRowCounts(Math.min(DEFAULT_ROW_COUNT, filterData({}).length));
+    });
+
+    it('will reset filters but not pagination and sorting', () => {
+      filterApply({ name: 'a' });
+      changePagination(PAGINATION_VALUES[0]);
+      cy.get(TOOLBAR)
+        .find(PAGINATION)
+        .find('button[data-action="next"]')
+        .then(($button) => {
+          cy.wrap($button).click();
+        });
+
+      cy.get('th[data-label="Name"]').find('button').click();
+      cy.get(TOOLBAR).find('button').contains('Reset filters').click();
+      cy.get(TOOLBAR).find(CHIP_GROUP).should('not.exist');
+      checkPaginationSelected(0);
+      checkCurrentPage(2);
+      cy.get('th[data-label="Name"]')
+        .should('have.attr', 'aria-sort')
+        .and('contain', 'ascending');
     });
 
     it('empty state is displayed when filters do not match any rule', () => {
--- a/src/Components/ClusterRules/ClusterRules.cy.js
+++ b/src/Components/ClusterRules/ClusterRules.cy.js
@@ -251,6 +251,17 @@ describe('cluster rules table', () => {
         .should('have.length', RULES_ENABLED);
     });
 
+    it('will reset filters but not pagination and sorting', () => {
+      filterApply({ description: 'Lo' });
+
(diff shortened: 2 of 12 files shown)


gkarat (commented):
I am afraid this feature is not relevant for all of the tables. Specifically, there is a scenario for the recs list table (/recommendations):

1. By default, you have two filters enabled: Clusters impacted: 1 or more, and Status: Enabled.
2. Remove the first filter, and you will get more recommendations available.
3. Navigate to for example page 3 and click Reset filters
4. You will get a blank table. This happens because the reset will show you the subset of rules that you saw with the removed default filter, and you don't reset pagination.

изображение

Can we make an exception for this table? All other tables seems to be working fine, since resetting to the default filters will always result in showing the same or larger number of items.

src/Services/Filters.js
```diff
@@ -53,6 +51,17 @@ const filtersInitialState = {
   clusterRulesState: CLUSTER_RULES_INITIAL_STATE,
 };
 
+export const resetFilters = (filters, initialState, updateFilters) => {
+  const { limit, offset, sortIndex, sortDirection } = filters;
+  updateFilters({
+    ...initialState,
+    ...(limit !== undefined && { limit }),
+    ...(limit !== undefined && { offset }),
```
gkarat, line 59: I think there is a typo?
```suggestion
    ...(offset !== undefined && { offset }),
```

src/Components/AffectedClustersTable/AffectedClustersTable.cy.js
```diff
@@ -523,6 +526,26 @@ describe('non-empty successful affected clusters table', () => {
...
+        .find(PAGINATION)
+        .find('button[data-action="next"]')
+        .then(($button) => {
+          cy.wrap($button).click();
+        });
+
+      cy.get('th[data-label="Name"]').find('button').click();
+      cy.get(TOOLBAR).find('button').contains('Reset filters').click();
+      cy.get(TOOLBAR).find(CHIP_GROUP).should('not.exist');
+      checkPaginationSelected('0');
```
gkarat, line 542: a nit-pick: since, in `checkPaginationSelected` you pass the argument to `eq`, you should pass a number, not string: https://docs.cypress.io/api/commands/eq#Arguments

```suggestion
      checkPaginationSelected(0);
```

src/Components/AffectedClustersTable/AffectedClustersTable.cy.js
```diff
@@ -523,6 +526,26 @@ describe('non-empty successful affected clusters table', () => {
...
+      checkCurrentPage('2', FOOTER);
```
gkarat, line 543: nit-pick: the same here: can we pass a number?
```suggestion
      checkCurrentPage(2, FOOTER);
```

src/Components/ClusterRules/ClusterRules.cy.js
```diff
@@ -251,6 +251,17 @@ describe('cluster rules table', () => {
         .should('have.length', RULES_ENABLED);
     });
 
+    it('will reset filters but not pagination and sorting', () => {
+      filterApply({ description: 'Lo' });
+
+      cy.get('th[data-label="Description"]').find('button').click();
+      cy.get(TOOLBAR).find('button').contains('Reset filters').click();
+      cy.get(TOOLBAR).find(CHIP_GROUP).should('not.exist');
+      cy.get('th[data-label="Description"]')
```
gkarat, line 260: you also can use `checkPaginationSelected` here to check the pagination (=can we check for pagination and sorting in tests for all pages?)

src/Services/Filters.js
johnsonm325: Nope, not a typo. The ClusterRules table, which doesn't have pagination, returns `0` for `offset` when it calls this `resetFilters` function. So, I check for `limit === undefined` because then I know I'm on the ClusterRules page.

src/Components/ClusterRules/ClusterRules.cy.js
johnsonm325: ClusterRules doesn't have pagination.

johnsonm325:
@gkarat this is ready for another review

gkarat (commented):
LGTM! Just one minor comment regarding `async` + we need to fix the failing test: it looks like if you increase the number of items matching the filter "Name: o", then it will help. Otherwise, the next button is disabled.

изображение

src/Components/RecsListTable/RecsListTable.js
```diff
@@ -110,8 +111,15 @@ const RecsListTable = ({ query }) => {
     filters.sortDirection,
   ]);
 
-  useEffect(() => {
-    setFilteredRows(buildFilteredRows(recs, filters));
+  useEffect(async () => {
```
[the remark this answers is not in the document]
johnsonm325: I added it and forgot to test it again later. It's not needed, so I removed it.

gkarat:
thanks Michael! looks good to me. and we can ignore the ci.int check: we need to update the test script in a separate PR

PR #451, opened 2022-11-11
feat(Tables): Align columns to fit their content

gkarat:
Just a small cosmetic change to make all tables align their columns so that they fit their content and don't have extra width.

Example from the affected clusters list:

изображение


mkholjuraev (approved):
LGTM! Nice to have in Patch as well :).

PR #453, opened 2022-11-11
[OCPADVISOR-41] Resolve pagination issues

gkarat:
Fixes https://issues.redhat.com/browse/OCPADVISOR-41.

After the table audit, there were couple of problems found in Advisor tables. This PR 1) sets default pagination to 50 for all pages, 2) removes extra margin for the bottom pagination at the affected clusters table.


mkholjuraev (approved):
LGTM! tested locally, the extra margin is removed, default row size is 50. Nice work! Tests need to be fixed though.

PR #464, opened 2022-11-22
Update frontend.yml

adamrdrew:
Changing the frontend path (which is used to create the openshift ingress routes) because it needs to match both the app name in package.json and the manifestLocation path so all of it hooks up together.


adamrdrew:
I think our PR check is always going to fail because we don't ship the deps in our build container to run cypress. Gonna need to huddle with my team on what to do about this.

PR #471, opened 2022-11-24
feat(clusters): adding regexp instead of semver

Fewwy:
https://issues.redhat.com/browse/OCPADVISOR-20
The discussion about "how to sort" clusters with prefixes is still open, so this PR is a draft

### DIFF
--- a/src/Components/Common/Tables.js
+++ b/src/Components/Common/Tables.js
@@ -1,7 +1,7 @@
 import capitalize from 'lodash/capitalize';
 import cloneDeep from 'lodash/cloneDeep';
 import { useEffect, useState } from 'react';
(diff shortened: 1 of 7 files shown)


src/Components/Common/Tables.js
```diff
@@ -222,8 +223,7 @@ export const updateSearchParams = (filters = {}, columnMapping) => {
 // TODO: move to Utils.js
 export const compareSemVer = (v1, v2, d) => d * compare(v1, v2);
```
Fewwy, line 224: This function should be updated if we want to add prefixes

gkarat:
Looks like this PR is no longer relevant.

PR #477, opened 2022-11-28
[OCPADVISOR-62] Reset page number on filter

johnsonm325:
On affected clusters and clusters list tables, if you change per page or reset/apply filters, you will stay on the same page. Pages should be reset to 1 if you change the per page count or apply or remove a filter. This PR fixes those issues.


gkarat (commented):
I tried to run it locally, but I see that at least for the Clusters list page, the sorting is reset after the click on Reset filters. I think this is not intended because you say only the current page should be reset to 1.

johnsonm325:
> I tried to run it locally, but I see that at least for the Clusters list page, the sorting is reset after the click on Reset filters. I think this is not intended because you say only the current page should be reset to 1.

I was not able to reproduce this.

gkarat:
Tried again, not sure what was the issue, but now I can't reproduce that too (probably was looking at the wrong env). LGTM 👍🏼

PR #497, opened 2023-01-13
refactor: Replace Main with section tag

gkarat:
This removes the usage of Main component that was connecting to the global chroming store and was prone to errors. The replacement is the section tag with custom PF class names.


Fewwy:
@gkarat LGTM

PR #498, opened 2023-01-16
feat(Single rec page): Show risk of change info

gkarat:
Implements https://issues.redhat.com/browse/OCPADVISOR-13.

This enables the risk of change (resolution risk) section at the single recommendation page.

## How to test

Run the PR locally and navigate to /openshift/insights/advisor/recommendations. Select any recommendation and move to the details page. The "Risk of change" header must be rendered with all the additional labels and information. You can also check the API response from `GET /rule/ID` endpoint and check the `resolution_risk` integer value corresponds to the correct label (Very low = 1, Low = 2, Moderate = 3, High = 4).

## Screenshots

image


src/Utilities/Helpers.js
```diff
@@ -0,0 +1,3 @@
+import React from 'react';
+
+export const strong = (str) => <strong>{str}</strong>;
```
gkarat, line 3: This was done because the webpack build for cypress behaved a bit differently. In the previous implementation, there was a circular dependency which cypress webpack config wasn't able to resolve, and `strong` was evaluated to `undefined`. It's tricky and I can't say how to properly fix it, but this module separation resolves that.

PR #506, opened 2023-01-19
refactor: Use useChrome hook to access chrome API

gkarat:
This removes all the occurrences of direct calls to insights.chrome and replaces it with the recommended useChrome hook. Also, the AMS API client is removed from the source code since the API is no longer utilized by Advisor.

## How to test:

There must be no visual differences nor change in the app behavior. All modules and pages must work as they should. For example, the page titles have to be updated (`chrome.updateDocumentTitle`), and navigation to other apps must not cause any errors (`chrome.onChromeEvent`, `unregister`)


Hyperkid123 (approved):
LGTM

Fewwy (approved):
Tested it in stage-beta and stage-stable env, LGTM

PR #507, opened 2023-01-19
feat(Recs list): Risk of change details section

gkarat:
Implements https://issues.redhat.com/browse/OCPADVISOR-66.

This adds a new section to expanded rule content on the recommendations list table. The section shows information about the risk of change for the given rule.

## How to test

Navigate to /openshift/insights/advisor/recommendations. Check that for the rows that have risk of change value, the expanded information contains the risk of change section. For the rules that do not have (currently, all external rules have risk of change, but the tests cover this case), the section must be empty in the expanded content.

## Screenshots

image


ikerreyes (approved):
I would have preferred a different approach for finding the right rows for testing. But given that it is small test suite and it is easy to track values it should be fine.
Thanks for the tests.

PR #514, opened 2023-01-26
Enable UI IQE tests as part of pr_checks

ikerreyes:
This MR enables the use of IQE CCX plugin tests for validating UI deployments.  CCXDEV-10136

It is just a PoC that it works. 

I am a bit concerned about having to rename the COMPONENT_NAME before running the tests and not being able to use `ocp-advisor` for that (`export COMPONENT_NAME="ccx-data-pipeline"`)

### DIFF
--- a/pr_check.sh
+++ b/pr_check.sh
@@ -18,20 +18,33 @@ COMMON_BUILDER=https://raw.githubusercontent.com/RedHatInsights/insights-fronten
 # --------------------------------------------
 IQE_PLUGINS="ccx"
 IQE_MARKER_EXPRESSION="ui"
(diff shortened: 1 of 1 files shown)


ikerreyes:
@gkarat @Fewwy I would like to hear you opinions

- if we see value in having these tests
- what to run as part of those tests (only smoke tests, all tests...) I think it is worth to limit them somehow
- which env do we want to use the backend services from (stage or prod)

Fewwy:
@ikerreyes Good questions, we can talk together next week and discuss how to better approach this :)

gkarat:
I'll close the PR. We can get back to it once it becomes relevant

PR #528, opened 2023-03-08
fix(OCPADVISOR-59): Fix not loading pages

gkarat:
Fixes https://issues.redhat.com/browse/OCPADVISOR-59.

This should fix the bug when the users navigated to the cluster or recommendation details page for the second and further times and couldn't see data loaded.

[Screencast from 03-08-2023 01:48:38 PM.webm](https://user-images.githubusercontent.com/31385370/223717362-1c77476e-bc54-403e-8416-f02ec337b06d.webm)

## How to test

Navigate to recommendation/id or cluster/id pages. Go back to the recommendations/clusters list table via breadcrumbs. Go again to any of the details pages. The end page should load fine (whereas for example in production it does not currently).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "@redhat-cloud-services/frontend-components-translations": "^3.2.4",
         "@redhat-cloud-services/frontend-components-utilities": "^3.3.13",
         "@redhat-cloud-services/rule-components": "^3.2.6",
-        "@reduxjs/toolkit": "^1.9.0",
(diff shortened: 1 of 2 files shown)


gkarat:
Some more details on the bug: looks like https://github.com/reduxjs/redux-toolkit/releases/tag/v1.9.0 has introduced a change which breaks our pages. Downgrading to 1.8.6 helps to get back things to normal, however would be nice to make investigation on what exactly started to cause the proble,

PR #542, opened 2023-04-06
Cypress setup improvements

gkarat:


config/cypress.webpack.config.js
```diff
@@ -11,6 +11,24 @@ plugins.push(
...
+      ),
+    },
+  },
+});
+
+webpackConfig.module.rules.push({
+  test: /cypress\/.*\.js$/,
+  exclude: /(node_modules|bower_components)/i,
+  use: ['babel-loader'],
+});
```
gkarat, line 30: This helps to enable JSX in `cypress/support/commands.js`

src/Components/Cluster/Cluster.cy.js
```diff
@@ -1,231 +1,188 @@
...
-        isSuccess: true,
-        data: singleClusterPageReport,
-      },
-      displayName: {
-        data: singleClusterPageReport.report.meta.cluster_name,
-      },
-      clusterId: 'foobar',
-    };
-  });
+// interceptors
```
gkarat, line 15: Intercepting network requests instead of mocking props helps to create more abstraction from the implementation details.

ikerreyes (approved):
Great idea @gkarat Do you have plans to expand this approach to other test files for consistency? Not now but in the future

Fewwy (approved):
Great job! I think we can use this as an example for future improvements in Cypress testing in our apps

PR #551, opened 2023-04-20
fix: Use correct empty state for 404 upgrade risks

gkarat:
Fixes https://issues.redhat.com/browse/OCPADVISOR-71.

When upgrade risks back end endpoint responds with 404:

- show "not available" message instead of the table,
- render alert banner above the tabs.

When upgrade risks back end endpoint responds with 503 and other error codes (different from 404):

- show "something went wrong" message instead of the table,
- do not render alert banner above the tabs.

## How to test

As it was described in previous upgrade risks PRs, deploy webpack server together with the mock service. Verify the behavior for clusters that will have 404 or other error codes.

### DIFF
--- a/cypress/utils/interceptors.js
+++ b/cypress/utils/interceptors.js
@@ -104,6 +104,14 @@ export const upgradeRisksInterceptors = {
         statusCode: 503,
       }
     ),
+  'error, not found': () =>
+    cy.intercept(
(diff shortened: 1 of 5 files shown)


Fewwy:
LGTM

Fewwy (approved):

PR #586, opened 2023-06-29
fix(OCPADVISOR-101): Rename "upgrade" risks to "update" risks

bastilian:
Implements https://issues.redhat.com/browse/OCPADVISOR-101.

This renames all occurances of "upgrade" to "update", even filenames and code occurances to avoid any confusion in the future wether it should be on or the other. 

Places where it interfaces with external sources are untouched, for example API calls, fixtures for them or feature flags and analytics, for now.


gkarat (approved):
LGTM! thanks a lot

PR #597, opened 2023-07-26
Minor changes for Upgrade Risk Predictions tests

ikerreyes:
Hi folks, I have been revisiting the upgrade risk related test to check if there was anything doubled in here an in IQE. I come across a nitpick, in case you see value in the fix: 

- Change folder (route) for upgrade-risk-predictions endpoint: Change the v1 to v2, to reflect the actual implementation


gkarat (approved):

gkarat:
Thank you! @ikerreyes

PR #614, opened 2023-09-13
fix(Risk of change): Change risk of change text

gkarat:
Implements https://issues.redhat.com/browse/OCPADVISOR-69.

This changes the description of different levels of risk of change.


leSamo:
Why was the text taken out of intl? Was this intentional soft deprecation or mistake?

gkarat:
@leSamo, yeah, I decided not to mess with it just because I needed to include ol, li tags. Given we want to deprecate it, I took a shortcut 😛

leSamo (approved):
Fair enough :D

PR #624, opened 2023-10-26
feat(ocpadvisor-121): folders routes and components

Fewwy:
added basic routes
added folders
added rough component files

### DIFF
--- a/src/Components/WorkloadsList/index.js
+++ b/src/Components/WorkloadsList/index.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import { useIntl } from 'react-intl';
+
+import PageHeader from '@redhat-cloud-services/frontend-components/PageHeader';
+
+import messages from '../../Messages';
+import WorkloadsListTable from '../WorkloadsListTable';
+import { Title } from '@patternfly/react-core';
+import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
+
+const WorkloadsList = () => {
+  const intl = useIntl();
+  const chrome = useChrome();
+
+  chrome.updateDocumentTitle(
+    intl.formatMessage(messages.documentTitle, {
+      subnav: intl.formatMessage(messages.clusters),
+    })
+  );
+
+  return (
+    <React.Fragment>
+      <PageHeader className="ins-c-clusters-header">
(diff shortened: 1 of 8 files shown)


src/Routes.js
```diff
@@ -80,6 +88,28 @@ export const AppRoutes = () => (
...
         }
       />
+      <Route
+        path="/workloads"
+        element={
+          <WorkloadsList
+            /**
+             * Generate random `key` to force component re-render,
+             * temporary workaround for https://issues.redhat.com/browse/OCPADVISOR-59
+             */ key={Math.random()}
```
gkarat, line 98: Can you please check if https://issues.redhat.com/browse/OCPADVISOR-59 is reproducible if we remove this `key` workaround now? If we have enough time, I would vote for investigating how this could be fixed properly (though if it's no longer reproducible, let's forget about it and remove `key` now).

src/Routes.js
```diff
@@ -80,6 +88,28 @@ export const AppRoutes = () => (
           />
         }
       />
+      <Route
```
gkarat, line 91: Is there a plan to use a feature flag for this?

src/Components/WorkloadsList/index.js
```diff
@@ -0,0 +1,37 @@
...
+    })
+  );
+
+  return (
+    <React.Fragment>
+      <PageHeader className="ins-c-clusters-header">
+        <Title headingLevel="h1" ouiaId="page-header">
+          {`${intl.formatMessage(messages.insightsHeader)} ${intl
+            .formatMessage(messages.clusters)
+            .toLowerCase()}`}
```
gkarat, line 27: Let's start soft-deprecating `intl` and don't use it in new components. Feel free to change it to plain text

Fewwy: Thanks, will do

src/Routes.js
Fewwy: The issue is still present. We can investigate it in a separate task, but it shouldn't be a part of the "structure setup PR"

src/Routes.js
Fewwy: Yes, forgot to add. Will do it. Thanks

AsToNlele (approved):
LGTM, thank you! :smile:

PR #626, opened 2023-10-30
feat(OCPADVISOR-123): filter configs filter slice and constants

Fewwy:
https://issues.redhat.com/browse/OCPADVISOR-123

### DIFF
--- a/src/Services/Filters.js
+++ b/src/Services/Filters.js
@@ -1,5 +1,8 @@
 import { createSlice } from '@reduxjs/toolkit';
-import { CLUSTERS_TABLE_CELL_LAST_SEEN } from '../AppConstants';
+import {
+  CLUSTERS_TABLE_CELL_LAST_SEEN,
+  WORKLOADS_TABLE_CELL_LAST_SEEN,
+} from '../AppConstants';
 
 // single recommendation page
 export const AFFECTED_CLUSTERS_INITIAL_STATE = {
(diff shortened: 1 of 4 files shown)


src/Services/Filters.js
```diff
@@ -44,11 +44,21 @@ export const CLUSTER_RULES_INITIAL_STATE = {
   text: '',
 };
 
+export const WORKLOADS_LIST_INITIAL_STATE = {
+  limit: 50,
+  offset: 0,
+  sortIndex: CLUSTERS_TABLE_CELL_LAST_SEEN,
```
gkarat, line 50: Should we use the constant for the workloads table instead?

gkarat (approved):
looks good to me! left a few comments

src/Services/Filters.js
gkarat: Why renaming? I think creating a new one is the right way since these are two separate tables.

AsToNlele (approved):
LGTM, good job! :smile:

src/Services/Filters.js
Fewwy: Aaaaaaaaah! You mean for the next table! Yes, there will be a constant for it as well, not in this PR though :D

Fewwy: I'm stupid...for some reason, I thought you were talking about the object itself...

AsToNlele (approved):
Good job! The filters look like in the mocks, should be straightforward to connect with the table. :smile:

PR #633, opened 2023-11-09
feat(OCPADVISOR-146): pagination for workloads table

Fewwy:
https://issues.redhat.com/browse/OCPADVISOR-146

Adds pagination to the workloads table

How to setup mock api:

Clone [insights-results-aggregator-mock](https://github.com/RedHatInsights/insights-results-aggregator-mock)
make build (takes a few minutes for me, coffee time ☕)
2.1. can speed it up with make build -j 4 to run multiple jobs at once
make run
Test the mocked api curl localhost:8080/api/insights-results-aggregator/v2/namespaces/dvo
Run the app MOCK=true npm run start:proxy:beta
Workloads should load mocked data

### DIFF
--- a/src/Components/WorkloadsListTable/WorkloadsListTable.js
+++ b/src/Components/WorkloadsListTable/WorkloadsListTable.js
@@ -51,7 +51,6 @@ const WorkloadsListTable = ({
   const [filteredRows, setFilteredRows] = useState([]);
   const [rowsFiltered, setRowsFiltered] = useState(false);
   const [filtersApplied, setFiltersApplied] = useState(false);
-  const [tempQuery, setTempQuery] = useState(0);
   const updateFilters = (payload) =>
     dispatch(updateWorkloadsListFilters(payload));
   const removeFilterParam = (param) =>
@@ -65,7 +64,6 @@ const WorkloadsListTable = ({
   useEffect(() => {
     setFilteredRows(buildFilteredRows(workloads));
   }, [
-    tempQuery,
     filters.namespace_name,
     filters.cluster_name,
     filters.general_severity,
@@ -76,8 +74,6 @@ const WorkloadsListTable = ({
 
   useEffect(() => {
     setRows(buildDisplayedRows(filteredRows));
-    //should be refactored to smth like setDisplayedRows(buildDisplayedRows(filteredRows));
-    //when we add pagination
     setRowsFiltered(true);
     setFiltersApplied(noFiltersApplied(filters).length > 0 ? true : false);
   }, [filteredRows, filters.limit, filters.offset]);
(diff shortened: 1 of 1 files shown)


src/Components/WorkloadsListTable/WorkloadsListTable.js
```diff
@@ -178,15 +187,28 @@ const WorkloadsListTable = ({
...
       <PrimaryToolbar
         pagination={{
           itemCount: data?.workloads.length || 0,
-          page: 1,
-          perPage: 20,
-          onSetPage: () => console.log('here should be a pagination'),
-          onPerPageSelect: () => console.log('here should be a pagination'),
+          page: page,
+          perPage: perPage,
+          onSetPage: onSetPage,
```
gkarat, line 210:
```suggestion
          page,
          perPage,
          onSetPage,
```

src/Components/WorkloadsListTable/WorkloadsListTable.js
```diff
@@ -235,10 +257,11 @@ const WorkloadsListTable = ({
...
         itemCount={data?.workloads.length || 0}
-        page={1}
-        perPage={20}
-        onSetPage={() => {}}
-        onPerPageSelect={() => {}}
+        page={page}
+        perPage={perPage}
+        onSetPage={onSetPage}
+        onPerPageSelect={onSetPerPage}
+        onPageInput={onSetPage}
```
gkarat, line 264: same here you can avoid duplicate names

src/Components/WorkloadsListTable/WorkloadsListTable.js
```diff
@@ -44,32 +44,28 @@ const WorkloadsListTable = ({
   //const workloads = data?.workloads || [];
   //to check all types of filters use the mockdata json
   const workloads = mockdata;
+  const perPage = filters.limit;
+  const page = filters.offset / filters.limit + 1;
```
gkarat, line 48: I feel we better have rounding here. Imagine offset set to 50 but limit to 15. the result page is 4.333333333333334. This scenario shouldn't happen if all of our pagination constants are divisible, but it can cause bugs theoretically with slightly other variations in the future..

src/Components/WorkloadsListTable/WorkloadsListTable.js
```diff
@@ -178,15 +187,28 @@ const WorkloadsListTable = ({
...
 
+  const onSetPage = (_e, pageNumber) => {
+    setRowsFiltered(false);
+    const newOffset = pageNumber * filters.limit - filters.limit;
+    updateFilters({ ...filters, offset: newOffset });
+  };
+
+  const onSetPerPage = (_e, perPage) => {
+    //THIS IS A DUMMY QUERY THAT WILL BE REMOVED WHEN WE WORK ON CONNECTING FILTERS/SORTING/PAGES TO THE URL PARAMS
+    setTempQuery(Math.random());
```
gkarat, line 198: You don't need this. In fact, the thing is that if you select the same per page value and it doesn't reload the page correctly is the indication of the bug in the current implementation and it has a chance to persist in the future.

The reason why it happens is that useEffect is not triggered because perPage has the same value (it is primitive and useEffect doesn't compare references like with objects, but just compares the value). Thus, setRowsFiltered(true) is not called neither. One of the safe ways to fix this is to not to call updateFilters in these callbacks if perPage or pageNumber are the same as currently applied values. The table will not go into the loading state and still show the same results which is correct.

gkarat (approved):
There are couple of things we need to improve I believe. But since it's an intermediate PR, it looks and works pretty much fine 👍🏼 thanks @Fewwy

src/Components/WorkloadsListTable/WorkloadsListTable.js
Fewwy: Fixed this bug

gkarat (approved):
lgtm! thanks @Fewwy

PR #634, opened 2023-11-13
feat(OCPADVISOR-128): workloads table sorting

Fewwy:
How to check the PR

Make sure that the columns are sortable

### DIFF
--- a/src/Components/WorkloadsListTable/WorkloadsListTable.js
+++ b/src/Components/WorkloadsListTable/WorkloadsListTable.js
@@ -10,7 +10,12 @@ import {
 import { PaginationVariant } from '@patternfly/react-core/dist/js/components/Pagination/Pagination';
 import {
   WORKLOADS_LIST_COLUMNS,
+  WORKLOADS_TABLE_CELL_LAST_SEEN,
+  WORKLOADS_TABLE_CELL_NAME,
+  WORKLOADS_TABLE_CELL_OBJECTS,
+  WORKLOADS_TABLE_CELL_RECOMMENDATIONS,
   WORKLOADS_TABLE_FILTER_CATEGORIES,
+  WORKLOADS_TABLE_SEVERITY,
 } from '../../AppConstants';
 import DateFormat from '@redhat-cloud-services/frontend-components/DateFormat';
 import { Link } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


src/Components/WorkloadsListTable/WorkloadsListTable.js
```diff
@@ -80,12 +85,49 @@ const WorkloadsListTable = ({
...
+          case WORKLOADS_TABLE_SEVERITY:
+            fst = Object.values(a.metadata.hits_by_severity || 0).reduce(
+              (a, b) => a + b,
+              0
+            );
+            snd = Object.values(b.metadata.hits_by_severity || 0).reduce(
+              (a, b) => a + b,
+              0
+            );
+            return fst > snd ? d : snd > fst ? -d : 0;
```
AsToNlele, line 118: You could do something like this

```suggestion
          case WORKLOADS_TABLE_SEVERITY:
            fst = a.metadata.hits_by_severity || { 1: 0, 2: 0, 3: 0, 4: 0 };
            snd = b.metadata.hits_by_severity || { 1: 0, 2: 0, 3: 0, 4: 0 };

            for (let i = 4; i > 0; i--) {
              if (fst[i] > snd[i]) return d;
              if (fst[i] < snd[i]) return -d;
            }

            return 0;
```

AsToNlele (changes requested):
Just to verify, was it confirmed the sorting for severity should be the same like in vulnerability? If that's the case I left a comment

src/Components/WorkloadsListTable/WorkloadsListTable.js
Fewwy: Ok, gonna do that :)

src/Components/WorkloadsListTable/WorkloadsListTable.js
```diff
@@ -98,8 +140,16 @@ const WorkloadsListTable = ({
...
-              <p key={`${index}-namespace`}>{item.namespace.name}</p>
+              {item.cluster.display_name ? (
+                <p key={`${index}-cluster`}>{item.cluster.display_name}</p>
+              ) : (
+                <p key={`${index}-cluster`}>{item.cluster.uuid}</p>
+              )}
+              {item.namespace.name ? (
+                <p key={`${index}-cluster`}>{item.namespace.name}</p>
+              ) : (
+                <p key={`${index}-cluster`}>{item.namespace.uuid}</p>
```
AsToNlele, line 151: Typo here, so we can get rid of the same key errors

```suggestion
                <p key={`${index}-namespace`}>{item.namespace.name}</p>
              ) : (
                <p key={`${index}-namespace`}>{item.namespace.uuid}</p>
```

Fewwy:
> Just to verify, was it confirmed the sorting for severity should be the same like in vulnerability? If that's the case I left a comment

Yeah, pretty much that's the behavior we can copy

AsToNlele (approved):
LGTM @Fewwy, thanks for the quick fix :smile:

PR #635, opened 2023-11-13
feat(OCPADVISOR-121): initial setup for cypress tests

Fewwy:
PR for the initial setup of Cypress for the workloads table


AsToNlele (approved):
thanks for the setup @Fewwy :smile:

PR #640, opened 2023-11-22
fix(OCPADVISOR-121): hotfix for workloads table

Fewwy:
We have to subscribe to data in useEffect, so the table would show data on the initial load.


AsToNlele (approved):
LGTM, ahh that slipped through :smile:

PR #642, opened 2023-11-22
test(Cypress): flaky test fix

AsToNlele:
just a test


Fewwy:
It seems this is not a proper solution to the issue we encountered with flaky tests

PR #643, opened 2023-11-23
test(OCPADVISOR-121): fixing flaky test with retries

Fewwy:
I added the object with retry parameters that indicate that this test will run more times if it fails.
{
        retries: {
          runMode: 2,
          openMode: 1,
        },
      }


bastilian:
Please, let's look into a proper fix. There must be something going wrong with the test in this setup.

Fewwy:
> Please, let's look into a proper fix. There must be something going wrong with the test in this setup.

Agree, that's just a test PR that I tried yesterday to see if it improves the situation.

PR #645, opened 2023-11-29
feat(OCPADVISOR-132): workloads recommendations expanded details

Fewwy:
https://issues.redhat.com/browse/OCPADVISOR-132
First prototype of the expanded details. 
Some of the data is missing from the API response and I had to "improvise" to add it
Also, according to the Shayna I don't need to reduce the width of the objects table in the expanded details.

### DIFF
--- a/src/Utilities/TemplateProcessor.js
+++ b/src/Utilities/TemplateProcessor.js
@@ -1,0 +1,76 @@
+import React from 'react';
+import doT from 'dot';
+import { marked } from 'marked';
+import sanitize, { simpleTransform } from 'sanitize-html';
+
+const DOT_SETTINGS = {
+  ...doT.templateSettings,
+  varname: 'pydata',
+  strip: false,
+};
+
+// codespan and code tags escape child text by default, therefore we have to remove all escapes from there
+const walkTokens = (token) => {
+  if (token.type === 'codespan' || token.type === 'code') {
+    token.text = token.text
+      .replace(/\\\*/g, '*')
+      .replace(/\\_/g, '_')
+      .replace(/\\~/g, '~');
+  }
+};
+
+marked.use({ walkTokens });
+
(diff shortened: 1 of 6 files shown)


src/Utilities/TemplateProcessor.js
```diff
@@ -0,0 +1,76 @@
...
+        ul: simpleTransform('ul', {
+          class: 'pf-v5-c-list',
+          style: 'font-size: inherit',
+        }),
+        a: simpleTransform('a', {
+          rel: 'noopener noreferrer',
+          target: '_blank',
+        }),
+      },
+    });
```
gkarat, line 53: Can we use TemplateProcessor component? This can be duplicating the component feature and implementation

src/Utilities/TemplateProcessor.js
```diff
@@ -0,0 +1,76 @@
...
+        dangerouslySetInnerHTML={{
+          __html: sanitized.replace(
+            /<\/a>/gim,
+            ` <i class="fas fa-external-link-alt"></i></a>` // add an icon to external links
+          ),
+        }}
+      />
+    );
+  } catch (error) {
+    console.warn(error, definitions, template);
```
gkarat, line 66: Since it's not a happy path and caused by an error while processing templates, we better log it on the error level. This will be also caught by Sentry.

```suggestion
    console.error(error, definitions, template);
```

Fewwy: Ok, thanks

src/Utilities/TemplateProcessor.js
Fewwy: Update, there is a chance that we won't have to use it at all if the Processing team will update their code for the fields

AsToNlele (approved):
Looks really good @Fewwy! :smile: 

Let's merge it and fix the padding between the icon and title later
image

PR #651, opened 2023-12-07
feat(rhineng5533): workloads recs filters

Fewwy:
https://issues.redhat.com/browse/RHINENG-5533
Workloads recommendations table filters
2 filters were implemented in this PR: Name/Description filter and object id filter
Name/description filter should filter rows by the name
Object id should filter data by the object id's (you can look up their values inside the expanded details)

I did some code cleanup as well - noFiltersApplied function is renamed and refactored so we could use the output of this function as a boolean to check if we should render "Reset filters" button or not.


src/Components/WorkloadRules/WorkloadRules.js
```diff
@@ -104,48 +134,158 @@ const WorkloadRules = ({ workload }) => {
...
+
+  const buildFilterChips = () => {
+    const localFilters = { ...filters };
+    delete localFilters.sortIndex;
+    delete localFilters.sortDirection;
+    return pruneFilters(localFilters, WORKLOAD_RULES_FILTER_CATEGORIES);
+  };
+
+  const activeFiltersConfig = {
+    showDeleteButton: filtersApplied ? true : false,
```
gkarat, line 264: Can we get rid of that? The toolbar should be able to decide whether to render the reset filters button or not internally. If the toolbar shows the reset button, then we most probably pass non-empty array to `filters`. If there are no filters applied, then `buildFiltersChips` should return an empty array so that the function is semantic.

src/Components/ExpandedRulesDetails.js/ExpandedRulesDetails.js
```diff
@@ -83,10 +82,13 @@ const ExpandedRulesDetails = ({ recommendations }) => {
             <strong>Note:</strong>
           </CardHeader>
           <CardBody>
-            Lorem ipsum dolor sit amet consectetur adipisicing elit. Numquam
-            nemo illo minima nam voluptatum tempora amet blanditiis velit
-            aliquid omnis, repudiandae reiciendis ipsam aperiam expedita eius
-            corrupti aliquam praesentium! Ullam?
+            SHOULD BE RENDERED ONLY WHEN THERE IS NO NAMES FOR CLUSTER AND
```
gkarat, line 85: Do we still need this caps-locked text?

src/Components/WorkloadRules/WorkloadRules.js
```diff
@@ -104,48 +134,158 @@ const WorkloadRules = ({ workload }) => {
...
-          {
-            title: <ExpandedRulesDetails recommendations={recommendations} />,
-          },
-        ],
-      },
-    ]);
+  const pruneFilters = (localFilters, filterCategories) => {
+    const prunedFilters = Object.entries(localFilters);
+    return prunedFilters.length > 0
+      ? prunedFilters.reduce((arr, item) => {
```
gkarat, line 190: nit-pick (for better readability):
```suggestion
      ? prunedFilters.reduce((arr, [name, value]) => {
```

src/Components/WorkloadRules/WorkloadRules.js
```diff
@@ -104,48 +134,158 @@ const WorkloadRules = ({ workload }) => {
...
-      },
-      {
-        cells: [
-          {
-            title: <ExpandedRulesDetails recommendations={recommendations} />,
-          },
-        ],
-      },
-    ]);
+  const pruneFilters = (localFilters, filterCategories) => {
```
gkarat, line 187: This function is huge. Any chance we can extract this to a separate module and not keep as a part of the WorkloadRules component? Having unit tests for it also would be very useful

gkarat (changes requested):
The total risk filter seems to be not working as expected (the description and object id filters are OK): 

https://github.com/RedHatInsights/ocp-advisor-frontend/assets/31385370/60277940-c774-4204-ae6a-aa5a4a49bcd4

Fewwy:
> The total risk filter seems to be not working as expected (the description and object id filters are OK):
> 
>  Screen.Recording.2024-01-04.at.11.50.15.mov

Total risk is not available in the API yet, will be implemented separately

src/Components/WorkloadRules/WorkloadRules.js
Fewwy: No we can't, we tried to make it work together with Sebastian and this is the solution we came up with

src/Components/WorkloadRules/WorkloadRules.js
```diff
@@ -40,35 +63,38 @@ const WorkloadRules = ({ workload }) => {
...
-        //   addFilterParam(FILTER_CATEGORIES.total_risk.urlParam, values),
+        key: `total_risk`,
+        onChange: (_e, values) => addFilterParam('total_risk', values),
         value: filters.total_risk,
-        items: FC.total_risk.values,
+        items: WORKLOAD_RULES_FILTER_CATEGORIES.total_risk.values,
       },
     },
     {
       label: 'object ID',
```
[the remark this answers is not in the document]
Fewwy: Done, I moved the filter configs to the app constants and made it a function that returns an array of configs

src/Components/WorkloadRules/WorkloadRules.js
Fewwy: Extracted, refactored it and wrote tests for it

gkarat (approved):
LGTM

PR #653, opened 2024-01-05
feat(RHINENG-5520): Workloads details sorting

Fewwy:
Jira link: https://issues.redhat.com/browse/RHINENG-5520
This PR covers sorting for the Description, Objects, and Modified columns. It doesn't cover the total risk column sorting because data is not available in the API yet.
When the first row is expanded and you click on the sort arrow it won't work. It is expected. This is gonna be fixed in a separate PR and covered by a separate Jira card.

How to check the PR 

1. Clone [insights-results-aggregator-mock](https://github.com/RedHatInsights/insights-results-aggregator-mock) or pull the latest changes
2. make build
3. can speed it up with make build -j 4 to run multiple jobs at once
4. make run
5. Test the mocked api curl localhost:8080/api/insights-results-aggregator/v2/namespaces/dvo
6. Run the app MOCK=true npm run start:proxy:beta
7. Workloads should load mocked data
8. Review the PR

### DIFF
--- a/src/Components/WorkloadRules/WorkloadRules.js
+++ b/src/Components/WorkloadRules/WorkloadRules.js
@@ -1,5 +1,6 @@
 import React, { useEffect, useState } from 'react';
 import {
+  SortByDirection,
   Table,
   TableBody,
   TableHeader,
@@ -76,10 +77,34 @@ const WorkloadRules = ({ workload }) => {
   );
 
   const buildDisplayedRows = (filteredRows, sortIndex, sortDirection) => {
-    void sortIndex;
-    void sortDirection;
+    const sortingRows =
+      sortIndex >= 0 && !expandFirst
+        ? [...filteredRows].sort((a, b) => {
+            const d = sortDirection === SortByDirection.asc ? 1 : -1;
+            const getValue = (item) => {
+              switch (sortIndex) {
+                case 1:
+                  return item[0].rule.details;
+                case 2:
+                  console.log('total risk sorting - we need data in the api');
+                  break;
+                case 3:
+                  return item[0].rule.objects.length;
+                case 4:
+                  return item[0].rule.modified;
+                default:
+                  return 0;
+              }
+            };
+            return getValue(a, sortIndex) > getValue(b, sortIndex)
+              ? d
+              : getValue(b, sortIndex) > getValue(a, sortIndex)
+              ? -d
+              : 0;
+          })
+        : [...filteredRows];
 
-    return filteredRows.flatMap((row, index) => {
+    return sortingRows.flatMap((row, index) => {
       const updatedRow = [...row];
       if (expandFirst && index === 0) {
         row[0].isOpen = true;
@@ -198,6 +223,16 @@ const WorkloadRules = ({ workload }) => {
     },
   };
 
+  const onSort = (_e, index, direction) => {
+    setRowsFiltered(false);
+    setExpandFirst(false);
+    return updateFilters({
+      ...filters,
+      sortIndex: index,
+      sortDirection: direction,
+    });
+  };
+
   return (
     <div id="workload-recs-list-table">
       <PrimaryToolbar
@@ -250,6 +285,11 @@ const WorkloadRules = ({ workload }) => {
         variant={TableVariant.compact}
         isStickyHeader
         canCollapseAll
+        sortBy={{
+          index: filters.sortIndex,
+          direction: filters.sortDirection,
+        }}
+        onSort={onSort}
       >
         <TableHeader />
         <TableBody />
--- a/src/Components/WorkloadRules/WorkloadRules.js
+++ b/src/Components/WorkloadRules/WorkloadRules.js
@@ -47,13 +47,17 @@ const WorkloadRules = ({ workload }) => {
   const [rowsFiltered, setRowsFiltered] = useState(false);
   const [filtersApplied, setFiltersApplied] = useState(false);
   const [expandFirst, setExpandFirst] = useState(true);
+  const [firstRule, setFirstRule] = useState('');
   const loadingState = isUninitialized || isFetching || !rowsFiltered;
   //FILTERS
   const filters = useSelector(({ filters }) => filters.workloadsRecsListState);
+
   const updateFilters = (payload) =>
     dispatch(updateWorkloadsRecsListFilters(payload));
+
   const addFilterParam = (param, values) => {
     setExpandFirst(false);
+    setFirstRule('');
     return _addFilterParam(filters, updateFilters, param, values);
   };
   const removeFilterParam = (param) =>
@@ -78,7 +82,7 @@ const WorkloadRules = ({ workload }) => {
 
   const buildDisplayedRows = (filteredRows, sortIndex, sortDirection) => {
     const sortingRows =
-      sortIndex >= 0 && !expandFirst
+      sortIndex >= 0 && !firstRule
         ? [...filteredRows].sort((a, b) => {
             const d = sortDirection === SortByDirection.asc ? 1 : -1;
             const getValue = (item) => {
@@ -226,6 +230,7 @@ const WorkloadRules = ({ workload }) => {
   const onSort = (_e, index, direction) => {
     setRowsFiltered(false);
     setExpandFirst(false);
+    setFirstRule('');
     return updateFilters({
       ...filters,
       sortIndex: index,
--- a/src/Components/WorkloadRules/WorkloadRules.js
+++ b/src/Components/WorkloadRules/WorkloadRules.js
@@ -1,6 +1,5 @@
 import React, { useEffect, useState } from 'react';
 import {
-  SortByDirection,
   Table,
   TableBody,
   TableHeader,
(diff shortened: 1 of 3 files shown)


src/Components/WorkloadRules/WorkloadRules.js
```diff
@@ -198,6 +227,17 @@ const WorkloadRules = ({ workload }) => {
     },
   };
 
+  const onSort = (_e, index, direction) => {
+    setRowsFiltered(false);
+    setExpandFirst(false);
+    setFirstRule('');
+    return updateFilters({
```
gkarat, line 234: `updateFilters` doesn't return anything. Let's just call the function.

```suggestion
    updateFilters({
```

src/Components/WorkloadRules/WorkloadRules.js
```diff
@@ -198,6 +227,17 @@ const WorkloadRules = ({ workload }) => {
     },
   };
 
+  const onSort = (_e, index, direction) => {
+    setRowsFiltered(false);
```
gkarat, line 231: Can you please clarify what we do with `rowsFiltered` in the component after we set it to false here? I can't find any place where the flag is set back to true (`setRowsFiltered(true)`). Does it mean that `loadingState` variable will be true?

gkarat (changes requested):
None of the columns work (even when the first row is not expanded):

https://github.com/RedHatInsights/ocp-advisor-frontend/assets/31385370/7bf77ab0-311e-4927-aea0-dcd206bc357b

The error stack:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'defaultDirection')
    at sortClicked (vendors-node_modules_patternfly_react-table_dist_esm_index_js.0c733fb8f751639cfe67.js:sourcemap:2969:40)
    at onClick (vendors-node_modules_patternfly_react-table_dist_esm_index_js.0c733fb8f751639cfe67.js:sourcemap:1288:329)
    at HTMLUnknownElement.callCallback (chrome-root.ef2664ff17126cff.js:25153:14)
    at Object.invokeGuardedCallbackDev (chrome-root.ef2664ff17126cff.js:25202:16)
    at invokeGuardedCallback (chrome-root.ef2664ff17126cff.js:25266:31)
    at invokeGuardedCallbackAndCatchFirstError (chrome-root.ef2664ff17126cff.js:25280:25)
    at executeDispatch (chrome-root.ef2664ff17126cff.js:30030:3)
    at processDispatchQueueItemsInOrder (chrome-root.ef2664ff17126cff.js:30062:7)
    at processDispatchQueue (chrome-root.ef2664ff17126cff.js:30075:5)
    at dispatchEventsForPlugins (chrome-root.ef2664ff17126cff.js:30086:3)
```

I suggest to have another look at the code and try to finish the sorting feature for all columns. There is honestly no need to open more intermediate PRs that partially implement the feature: let's just focus on this one and make sure we have the sorting enabled for this page.

gkarat:
For total risk: I see that we don't have the value coming from API yet. But, do we know what's the type of the value will be? String, number? Do we know the field from the response where the value will be available? I think we already have enough information to implement sorting for this column since we do sorting for this table on the client side.

gkarat:
Oopsie, the sorting works indeed fine 👍🏼 I think I was testing with a wrong commit. I would still like to see total risk column sorting implemented (see my last comment).

src/Components/WorkloadRules/WorkloadRules.js
Fewwy: Regarding that, we setting it to true on the line 76

src/Components/WorkloadRules/WorkloadRules.js
gkarat: It's not removed in the last commit

src/Utilities/Workloads.js
```diff
@@ -144,3 +145,48 @@ export const pruneWorkloadsRulesFilters = (localFilters, filterCategories) => {
...
   }, []);
 };
+
+export const switchSort = (sortIndex, item) => {
+  const rule = item[0]?.rule; // Using optional chaining to check if 'rule' is defined
+  switch (sortIndex) {
+    case 1:
+      return rule?.details || '';
+    case 2:
+      return rule?.total_risk || [];
```
gkarat, line 155: Why [] was selected as a default value for the number-type value?

src/Utilities/Workloads.js
```diff
@@ -144,3 +145,48 @@ export const pruneWorkloadsRulesFilters = (localFilters, filterCategories) => {
     return arr;
   }, []);
 };
+
+export const switchSort = (sortIndex, item) => {
+  const rule = item[0]?.rule; // Using optional chaining to check if 'rule' is defined
```
gkarat, line 150: nit-pick: what are the circumstances when the rule can be not defined? are they expected/anticipated? if not, then there is no reason to introduce optional chaining.

src/Utilities/Workloads.js
```diff
@@ -144,3 +145,48 @@ export const pruneWorkloadsRulesFilters = (localFilters, filterCategories) => {
...
+  }
+};
+
+export const sortWithSwitch = (
+  sortIndex,
+  sortDirection,
+  filteredRows,
+  firstRule
+) => {
+  return sortIndex >= 0 && !firstRule
```
gkarat, line 171: nit-pick: can we remove the `firstRule` condition from this function? so that it doesn't complicate the logic here and more adhere to its semantics. instead, you can move this condition to WorkloadsRules component https://github.com/RedHatInsights/ocp-advisor-frontend/pull/653/commits/77277516615d3aba5df120fd2123c6ff890d8f96#diff-ac9d86077d652e9ca2918aa706320d2fff4adf9fb67d2d42752b639a67d3fb18R85: if firstRule is truthy, then don't call sortWithSwitch on the filteredRows.

gkarat, line 171: Looking at switchSort: it looks like the sorting index starts from 1. I think we need to rewrite it to
```
  return sortIndex >= 1 && !firstRule
```
instead, WDYT?

src/Utilities/Workloads.js
```diff
@@ -144,3 +145,48 @@ export const pruneWorkloadsRulesFilters = (localFilters, filterCategories) => {
...
+    case 1:
+      return rule?.details || '';
+    case 2:
+      return rule?.total_risk || [];
+    case 3:
+      return rule?.objects?.length || 0;
+    case 4:
+      return rule?.modified || '';
+    default:
+      return 0;
```
gkarat, line 161: This should at least console.error about the issue: it's not expected that the value is less than 1 or more than 4. The unexpected behaviour is an error case.

src/Components/WorkloadRules/WorkloadsRules.test.js
```diff
@@ -75,3 +77,73 @@ describe('pruneWorkloadsRulesFilters function', () => {
...
+
+  test('should sort rows in descending order based on switchSort', () => {
+    const filteredRows = [
+      { rule: { details: 'abc' } },
+      { rule: { details: 'xyz' } },
+    ];
+    const result = sortWithSwitch(1, SortByDirection.desc, filteredRows, false);
+    expect(result).toEqual([
+      { rule: { details: 'abc' } },
+      { rule: { details: 'xyz' } },
```
gkarat, line 146: Shouldn't this be an opposite? If we sort in the **descending** order...

src/Components/WorkloadRules/WorkloadRules.js
Fewwy: That's a mistake on my end, forgot to add this change while I was jumping between 2 branches. I am pushing new update right now

gkarat (commented):
Some things can be improved and polished, let's take another round on this PR. Otherwise, LGTM!

src/Utilities/Workloads.js
Fewwy: I tried to remove this condition and see how it works without it. Works fine without !firstrule

Fewwy: Updated it

src/Components/WorkloadRules/WorkloadsRules.test.js
Fewwy: Yep, I updated the tests

src/Utilities/Workloads.js
Fewwy: Yes, for some reason I automatically added default while its not necessary at all, remove it

src/Utilities/Workloads.js
Fewwy: Mistake, updated it to a string

src/Utilities/Workloads.js
```diff
@@ -147,29 +147,22 @@ export const pruneWorkloadsRulesFilters = (localFilters, filterCategories) => {
...
   switch (sortIndex) {
     case 1:
       return rule?.details || '';
     case 2:
-      return rule?.total_risk || [];
+      return rule?.total_risk || '';
     case 3:
       return rule?.objects?.length || 0;
     case 4:
       return rule?.modified || '';
```
gkarat, line 159: I am not sure it makes sense to use default values at all here. All of the fields are expected to be present in the response, is it correct? In that case we have to remove the optional chaining and also the default values. Anything odd happening: we should have a proper error, not a value which is guessed and replaced as a placeholder.

```
    case 1:
      return rule.details;
    case 2:
      return rule.total_risk;
    case 3:
      return rule.objects.length;
    case 4:
      return rule.modified;
```

gkarat (commented):
LGTM! Just one more thing and we are ready to go

src/Utilities/Workloads.js
Fewwy: Ok, replacing that

gkarat (approved):
LGTM! Thanks for all the edits

PR #655, opened 2024-01-09
feat(RHINENG-6801): Workloads rules url params & filters

Fewwy:
This PR is dependent on the https://github.com/RedHatInsights/ocp-advisor-frontend/pull/653
otherwise, it's hard to make URL params work for the sorting :) 

TOTAL RISK is not present in the API yet, so filtering and sorting by this parameter won't work

There is an issue that for some reason I get offset applied to the URL so I have to fix that as well

How to check the PR
Check that filters and sorting are correctly added to the URL params.

How to run it locally:
1. Clone [insights-results-aggregator-mock](https://github.com/RedHatInsights/insights-results-aggregator-mock) or pull the latest changes
2. make build
3. can speed it up with make build -j 4 to run multiple jobs at once
4. make run
5. Test the mocked api curl localhost:8080/api/insights-results-aggregator/v2/namespaces/dvo
6. Run the app MOCK=true npm run start:proxy:beta
7. Workloads should load mocked data
8. Review the PR

### DIFF
--- a/src/Components/WorkloadRules/WorkloadRules.js
+++ b/src/Components/WorkloadRules/WorkloadRules.js
@@ -1,5 +1,6 @@
 import React, { useEffect, useState } from 'react';
 import {
+  SortByDirection,
   Table,
   TableBody,
   TableHeader,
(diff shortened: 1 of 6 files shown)


src/Components/WorkloadRules/WorkloadRules.js
```diff
@@ -57,10 +64,10 @@ const WorkloadRules = ({ workload }) => {
 
   const addFilterParam = (param, values) => {
     setExpandFirst(false);
-    return _addFilterParam(filters, updateFilters, param, values);
+    return workloadsRulesAddFilterParam(filters, updateFilters, param, values);
```
gkarat, line 67: Probably, we can remove return here?

```suggestion
    workloadsRulesAddFilterParam(filters, updateFilters, param, values);
```

gkarat (approved):
LGTM! Didn't spot any problems while testing locally, and code-wise looks everything OK

PR #662, opened 2024-01-24
fix(rhineng-5493): breadcrumbs fix

Fewwy:
Breadcrumbs fix, because of the Double name for the workloads that contain BOTH cluster ID and Namespace ID - breadcrumbs component had to adjusted.
I wrote tests to check that it is fixed
https://issues.redhat.com/browse/RHINENG-5493


AsToNlele (approved):
Thanks for the fix, LGTM :smile:

PR #663, opened 2024-01-24
Fix workloads table empty state

Fewwy:
New empty state (I don't have mocks for it ready but we discussed it with Shanya personally)
Should appear in case when there are no search results
image


AsToNlele (approved):
Works as described, LGTM! :smile:

PR #666, opened 2024-01-29
fix(workloads): fixing scrollbar and close button overlap

Fewwy:
Previously:
image

After fix:
image


AsToNlele (approved):
LGTM, much better now! :smile:

PR #670, opened 2024-02-05
fix(rhineng-7889): fixed a filter bug and wrote more tests

Fewwy:
I had to rethink how I map filters for the workload details.
I made a function using lodash methods to omit any empty filters and create a new object named cleanedUpFilters
After it is created we can check if we have matching recommendations based on the new object pairs of filterKey and filterValue.

I added basic tests to check that it works as expected.
Also, I moved this function to the Workloads utils file


AsToNlele (approved):
LGTM, Good job @Fewwy :smile:

PR #673, opened 2024-02-09
feat(rhineng-7974): severity link with path and filters

Fewwy:
Severity links now lead to the workload details with the applied filter.
Critical severity equals to critical filter applied and same for the rest of severities


AsToNlele (approved):
Works as described, LGTM :smile:

PR #674, opened 2024-02-09
feat(rhineng-8063): expanded row behavior and more mock data variety

Fewwy:
The behavior of row expanded should be the same as in cluster details.
To test:
Try switching expanded rows on and off and apply filters. The expanded row should stay expanded if it is shown in the filtered results


AsToNlele (approved):
Good job! :smile:

PR #675, opened 2024-02-13
feat(RHINENG-7886): add a beta check for the prod preview release

Fewwy:
To make a production preview release possible I need to add a beta check for the workload routes. That would allow us to turn the feature flag in production ON and serve workloads only in the production preview
I already prepared a PR for the chrome navigation https://github.com/RedHatInsights/chrome-service-backend/pull/403

beta off
image
beta on 
image

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -12,6 +12,7 @@ import useFeatureFlag, {
   WORKLOADS_ENABLE_FLAG,
 } from './Utilities/useFeatureFlag';
 import { ErrorState } from './Components/MessageState/EmptyStates';
+import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
 
 const Cluster = lazy(() =>
   import(/* webpackChunkName: "ClusterDetails" */ './Components/Cluster')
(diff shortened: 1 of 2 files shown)


src/Routes.js
```diff
@@ -97,7 +100,7 @@ export const AppRoutes = () => {
         <Route
           path="/workloads"
           element={
-            workloadsEnabled ? (
+            workloadsEnabled && beta ? (
```
gkarat, line 103: This will prevent the workloads feature from working in stage-stable environment. Is it what we are aiming to do?

Fewwy: Yes, this is going to be inevitable if we want to do a production preview release without showing it in prod stable :(

Fewwy: The plan is to remove feature flag and these checks as soon as we release to prod stable

gkarat: > Yes, this is going to be inevitable if we want to do a production preview release without showing it in prod stable :(

I believe this is doable with:
```
!(chrome.isProd() && chrome.isBeta()) && workloadsEnabled
```

gkarat (approved):

PR #681, opened 2024-02-21
tests: Upgrade Cypress tests to support PF5 and React 18

AsToNlele:
Sorry for taking so long but it was a drive,

Most of the Cypress versions hanged in Travis. Found that `12.3.0` works well so I pinned that version.

Some utils from the `frontend-components/utils` package weren't fully compatible, so some tests import the local utils.

There was an issue with Breadcrumbs tests, where clicking the link would break Cypress (most likely related to the `cypress/react18` package), we removed the cypress test and Sasha replaced it with an RTL test [here](https://github.com/RedHatInsights/ocp-advisor-frontend/pull/683)

Will squash after review


AsToNlele:
All cypress tests except `Breadcrumbs.cy.js` are fixed. I'm facing an issue with relative links. The breadcrumb always sets the `href` to `/`. My guess is that it uses the url from cypress instead the one provided in the `MemoryRouter` couldn't find any fix.

AsToNlele:
Travis timeouts bruh

AsToNlele:
"cypress": "12.3.0" failed only one test, will try the latest version + retries

AsToNlele:
can't believe it, tests passed 😭

Fewwy (approved):
LGTM

PR #698, opened 2024-03-04
Connect real api

Fewwy:
Connecting real API to the UI

Removed mocked data
Added a conditional rendering to hide objects table and button when the objects are absent for the recommendation.
Passed more data to the Expanded Rules Details to render the data using template processor
Conditional rendering to hide the Note part of the Expanded Rules Details component. It is dependent on the existence of a namespace Name.

### DIFF
--- a/src/Components/ExpandedRulesDetails.js/ExpandedRulesDetails.js
+++ b/src/Components/ExpandedRulesDetails.js/ExpandedRulesDetails.js
@@ -29,8 +29,9 @@ const ExpandedRulesDetails = ({
   more_info,
   resolution,
   objects,
-  definitions,
   namespaceName,
+  reason,
+  extra_data,
 }) => {
   const objectsArePresent = !_.isEmpty(objects);
   const [objectsModalOpen, setObjectsModalOpen] = useState(false);
@@ -56,8 +57,7 @@ const ExpandedRulesDetails = ({
                 <strong>Detected issues</strong>
               </CardHeader>
               <CardBody>
-                This should be a reason field and extradata should provide us an
-                array of reasons to list here
+                <TemplateProcessor template={reason} definitions={extra_data} />
               </CardBody>
             </Card>
           </StackItem>
@@ -73,7 +73,7 @@ const ExpandedRulesDetails = ({
               <CardBody>
                 <TemplateProcessor
                   template={resolution}
-                  definitions={definitions}
+                  definitions={extra_data}
                 />
               </CardBody>
             </Card>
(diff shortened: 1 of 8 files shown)


src/Components/ExpandedRulesDetails.js/ExpandedRulesDetails.js
```diff
@@ -20,11 +20,20 @@ import PropTypes from 'prop-types';
 import TemplateProcessor from '@redhat-cloud-services/frontend-components-advisor-components/TemplateProcessor/TemplateProcessor';
 import ObjectsModal from '../ObjectsModal/ObjectsModal';
 import { ObjectsTableColumns } from '../../AppConstants';
+import _ from 'lodash';
```
mkholjuraev, line 23:
```suggestion
import isEmpty from 'lodash/isEmpty';
```

This helps to load only what is needed.

src/Components/WorkloadRules/WorkloadRules.js
```diff
@@ -143,6 +143,7 @@ const WorkloadRules = ({ workload }) => {
   };
 
   const buildFilteredRows = (allRows, filters) => {
+    console.log(allRows, 'allrows');
```
mkholjuraev, line 146: We can remove this, I believe

mkholjuraev (commented):
I have run the app, I could see that now real API is used and there are no issues related to it. Let's fix the tests and I will have another quick look and probably we can proceed.

src/Components/WorkloadRules/WorkloadRules.js
Fewwy: Yes, forgot to delete that

src/Components/ExpandedRulesDetails.js/ExpandedRulesDetails.js
```diff
@@ -20,11 +20,20 @@ import PropTypes from 'prop-types';
...
-const ExpandedRulesDetails = ({ more_info, resolution, objects }) => {
+const ExpandedRulesDetails = ({
+  more_info,
+  resolution,
+  objects,
+  namespaceName,
+  reason,
+  extra_data,
+}) => {
+  const objectsArePresent = !isEmpty(objects);
```
AsToNlele, line 36: Just a small note, maybe we could make the bundle a bit smaller by removing the lodash function. I believe it always an array so we could check it like this
```suggestion
  const objectsArePresent = Array.isArray(objects) && objects.length > 0;
```

AsToNlele (approved):
Works as expected, good job! :smile: 
Left a small note

PR #700, opened 2024-03-05
fix(RHINENG-5526): More conditional rendering for expanded details and fixing component structure

Fewwy:
Mocks:
https://www.sketch.com/s/46f6d8e3-a4d0-4249-9d57-e6a79b518a6d/a/j4Pa8La
Added a check to render Additional info only in case when more info is present

restructured Steps to the Resolve section. The table was outside the "card" component. So I made it one card component and one Stack Item so it wouldn't have issues with margins

Before
image

After
image

I added codeblock component from Alex PR to this PR, its faster that way and we won't have to solve conflicts
image

### DIFF
--- a/src/Components/ExpandedRulesDetails.js/ExpandedRulesDetails.js
+++ b/src/Components/ExpandedRulesDetails.js/ExpandedRulesDetails.js
@@ -62,7 +62,6 @@ const ExpandedRulesDetails = ({
               </CardBody>
             </Card>
           </StackItem>
-
           {objectsArePresent && (
             <React.Fragment>
               <Divider />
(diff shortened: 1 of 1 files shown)


gkarat (approved):
The change looks good, thanks @Fewwy. Though, I am worried we don't have any unit tests for these components (ExpandedRulesDetails, OpenShiftCodeBlocks, etc.). These are huge chunks of code that we need to cover ideally now.

If you have capacity, please make sure the components are tested, or do it in a separate PR.

gkarat:
Also, none of the commits reference the ticket. It's a good practice to mention the tickets so that your PR are visible in the Jira fields

PR #703, opened 2024-03-08
fix(RHINENG-5526): empty state for 404 error

Fewwy:
Now we're showing a different empty state on the 404 error. 
https://www.sketch.com/s/46f6d8e3-a4d0-4249-9d57-e6a79b518a6d/a/P1Wm4yQ


AsToNlele (approved):
Good point, well done! :smile:

PR #704, opened 2024-03-08
fix(rhineng-5526): moving note text into header to align with mocks

Fewwy:
Note: 
should be inside the text message and not separate
like on mocks
image


AsToNlele (approved):
Much better, thanks! :smile:

PR #706, opened 2024-03-13
feat(RHINENG-8253): Update risk label in cluster page

AsToNlele:
[RHINENG-8253](https://issues.redhat.com/browse/RHINENG-8253)

### DIFF
--- a/cypress.config.js
+++ b/cypress.config.js
@@ -18,5 +18,4 @@ module.exports = defineConfig({
       return config;
     },
   },
-  retries: 2,
 });
--- a/cypress/utils/table.js
+++ b/cypress/utils/table.js
@@ -88,7 +88,9 @@ function checkFiltering(
     checkEmptyState(emptyStateTitle);
     checkTableHeaders(tableHeaders);
   } else {
+    cy.get('[data-ouia-component-id=loading-skeleton]').should('not.exist');
     cy.get(`td[data-label="${columnName}"]`)
+      .should('have.length', values.length)
       .then(($els) => {
         return _.map(Cypress.$.makeArray($els), 'innerText');
       })
@@ -197,7 +199,10 @@ function checkSorting(
     ),
     dataField
   );
+  cy.get('[data-ouia-component-id=loading-skeleton]').should('not.exist');
+  const expectedLength = sortedValues.slice(0, nExpectedRows).length;
   cy.get(`td[data-label="${columnField}"]`)
+    .should('have.length', expectedLength)
     .then(($els) => {
       return _.map(Cypress.$.makeArray($els), 'innerText');
     })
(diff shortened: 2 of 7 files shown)


src/Components/ClustersListTable/ClustersListTable.js
```diff
@@ -144,14 +165,24 @@ const ClustersListTable = ({
         );
       }
       const ver = toValidSemVer(it.cluster_version);
+      const upgrade =
+        upgradeArr.length > 0 ? upgradeArr[index]?.upgrade_recommended : null;
```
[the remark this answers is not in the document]
AsToNlele: I got a suggestion [here](https://redhat-internal.slack.com/archives/CLABA9CHY/p1709821030450009?thread_ts=1707984317.756049&cid=CLABA9CHY), but I'll doublecheck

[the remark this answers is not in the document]
AsToNlele: Verified [here](https://redhat-internal.slack.com/archives/CLABA9CHY/p1710855873495829?thread_ts=1707984317.756049&cid=CLABA9CHY)

AsToNlele:
I've restarted like a million times all is working...

Fewwy (approved):

cypress/utils/table.js
```diff
@@ -89,6 +89,8 @@ function checkFiltering(
     checkTableHeaders(tableHeaders);
   } else {
     cy.get('[data-ouia-component-id=loading-skeleton]').should('not.exist');
+    // eslint-disable cypress/no-unnecessary-waiting
+    cy.wait(1000);
```
gkarat, line 93: I think this is a very good example of where we actually don't need unnecessary waiting :-)

Can we think of why our tests are timing out or not able to find the expected data under the columns? Is it the fact that the table is still in the loading state? If so, what should appear if we consider the table loaded? Instead of waiting generic 1 second, can we just wait for some of the first row data to appear?

cypress.config.js
```diff
@@ -18,4 +18,5 @@ module.exports = defineConfig({
       return config;
     },
   },
+  retries: 2,
```
gkarat, line 21: I see this was introduced earlier already on master. But I believe this is a general anti-pattern when dealing with flaky tests. I strongly discourage from using any retries since we will have even less insights into why our tests or our UI is flaky.

gkarat (changes requested):

cypress/utils/table.js
AsToNlele: Yeah I wanted to give it a try if it eventually helps something, looks like it doesn't

AsToNlele: Loading state I've fixed few lines below by adding this
`cy.get('[data-ouia-component-id=loading-skeleton]').should('not.exist');`
Waits until the skeleton is gone

cypress.config.js
AsToNlele: In the end this usually didn't help with anything, if it failed once then it failed in the retries as well. But I agree it shouldn't be there

gkarat (approved):

PR #707, opened 2024-03-13
fix(bug): remove redundant API call

Fewwy:
I missed the fact that we're making the API call twice, I removed the redundant one and passed the data from the parent component


gkarat (commented):
I have tested the PR and the number of requests has been indeed reduced. Although, I can see that there is something wrong with the "Cluster ID" filed in the header, it shows the ID of a namespace, not cluster:

Screenshot 2024-03-14 at 17 29 48

gkarat (approved):
LGTM! Thanks @Fewwy

PR #717, opened 2024-03-20
fix(tests): fixed workload details tests after pf update

Fewwy:
fixed workload details tests after pf update


AsToNlele (approved):
Works for me :smile:

PR #718, opened 2024-03-20
chore(deps): updating packages with audit

Fewwy:


AsToNlele:
afaik the config package updates requires more stuff to change, e.g. adding the `fec.config.js`

PR #721, opened 2024-03-26
fix(RHINENG-8253): Update risk label condition

AsToNlele:
[RHINENG-8253](https://issues.redhat.com/browse/RHINENG-8253)

There was a miscommunication when the `Update risk` label should be displayed

When the `upgrade-risks-prediction` endpoint returns a prediction where `"upgrade_recommended": false`, then we display the `Update risk` label for that cluster.

How to test:
1. Run the app in prod beta `npm run start`, select prod, then beta
2. Go to Clusters
3. There should be a system which has the label
image

4. Verify it's the correct system by checking the endpoint

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@redhat-cloud-services/frontend-components-charts": "^3.2.6",
     "@redhat-cloud-services/frontend-components-notifications": "^4.1.0",
     "@redhat-cloud-services/frontend-components-translations": "^3.2.7",
-    "@redhat-cloud-services/frontend-components-utilities": "^4.0.7",
+    "@redhat-cloud-services/frontend-components-utilities": "^4.0.9",
     "@reduxjs/toolkit": "^2.2.0",
     "@unleash/proxy-client-react": "^4.1.2",
     "axios": "^1.6.7",
--- a/src/Components/ClustersListTable/ClustersListTable.js
+++ b/src/Components/ClustersListTable/ClustersListTable.js
@@ -53,7 +53,7 @@ import {
 } from '../MessageState/EmptyStates';
 import { coerce } from 'semver';
 import { BASE_PATH } from '../../Routes';
-import { useLazyGetClustersUpgradeRisksQuery } from '../../Services/SmartProxy';
+import { useAxiosWithPlatformInterceptors } from '@redhat-cloud-services/frontend-components-utilities/interceptors';
 
 const ClustersListTable = ({
   query: { isError, isUninitialized, isFetching, isSuccess, data, refetch },
@@ -63,7 +63,6 @@ const ClustersListTable = ({
   const updateFilters = (payload) =>
     dispatch(updateClustersListFilters(payload));
   const filters = useSelector(({ filters }) => filters.clustersListState);
-  const [getClustersUpgradeRisks] = useLazyGetClustersUpgradeRisksQuery();
 
   const clusters = data?.data || [];
   const page = filters.offset / filters.limit + 1;
@@ -79,8 +78,10 @@ const ClustersListTable = ({
   const successState = isSuccess;
   const noMatch =
     clusters.length > 0 &&
-    filteredRows.length === 0 &&
-    displayedRows.length === 0;
+    filteredRows?.length === 0 &&
+    displayedRows?.length === 0;
+
+  const axios = useAxiosWithPlatformInterceptors();
 
   const removeFilterParam = (param) =>
     _removeFilterParam(filters, updateFilters, param);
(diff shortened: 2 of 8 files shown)


src/Components/ClustersListTable/ClustersListTable.js
```diff
@@ -137,23 +145,32 @@ const ClustersListTable = ({
...
-        }).unwrap();
-        if (upgradeRisks.status === 'ok') {
-          upgradeArr = upgradeRisks.predictions;
+        const res = await axios.post(
+          '/api/insights-results-aggregator/v2/upgrade-risks-prediction',
+          { clusters: clusterArr },
+          {
+            signal,
+          }
+        );
```
AsToNlele, line 162: Removed the lazy query and replaced it with a bare axios call, needed to insert the signal which is used to cancel the request

src/Components/ClustersListTable/ClustersListTable.js
```diff
@@ -137,23 +145,32 @@ const ClustersListTable = ({
...
     const clusterArr = filtered.map((cluster) => cluster.cluster_id);
     let upgradeArr = [];
     if (clusterArr.length > 0) {
       try {
-        const upgradeRisks = await getClustersUpgradeRisks({
-          clusters: clusterArr,
-        }).unwrap();
-        if (upgradeRisks.status === 'ok') {
-          upgradeArr = upgradeRisks.predictions;
+        const res = await axios.post(
```
mkholjuraev, line 156: Let's use the shared useAxiosWithPlatformInterceptors for better error handling. Refer: https://redhat-internal.slack.com/archives/C02T90PLJ5V/p1711451779574089.

AsToNlele: I'm unable to cancel the request with this hook, looks like its a different instance on each request. I've tried to use the hook and importing the `instance` from `/interceptors` as well.

Fewwy:
When I am trying to test this PR against prod stable I get this 
image

AsToNlele:
Thanks, will try with more systems

AsToNlele:
Will use the `useFetchBatched` hook from the utils package when it's released, I want to test in the CI.

AsToNlele:
Just realized, if a user had thousands of systems, it would request the UpgradeRisk for all of them, because we do the pagination/sorting as the last step. This page will need a refactor

src/Components/ClustersListTable/ClustersListTable.js
```diff
@@ -137,27 +148,88 @@ const ClustersListTable = ({
...
+          it.cluster_version,
+          it.total_hit_count,
+          it.hits_by_total_risk?.[4] || 0,
+          it.hits_by_total_risk?.[3] || 0,
+          it.hits_by_total_risk?.[2] || 0,
+          it.hits_by_total_risk?.[1] || 0,
+          it.last_checked_at,
+        ],
+      };
+    });
```
AsToNlele, line 170: Not completely happy about this, you could say we're doing it once here and later with the labels, but the data needs to be properly sorted, so we don't need to request additional data.

AsToNlele:
Since we're now capped at 100 clusters by the pagination, we don't need to batch requests.

AsToNlele:
During cypress testing I've noticed sometimes we get an `req is not defined` error. This is an [issue](https://github.com/axios/axios/issues/6301) in the axios library. The frontend-components-utils package has to either use upgrade to the `1.x` version or not use the `useAxiosWithInterceptors` hook at all.

package.json
```diff
@@ -36,11 +36,12 @@
     "@redhat-cloud-services/frontend-components-charts": "^3.2.6",
     "@redhat-cloud-services/frontend-components-notifications": "^4.1.0",
     "@redhat-cloud-services/frontend-components-translations": "^3.2.7",
-    "@redhat-cloud-services/frontend-components-utilities": "^4.0.7",
+    "@redhat-cloud-services/frontend-components-utilities": "^4.0.9",
     "@reduxjs/toolkit": "^2.2.0",
     "@unleash/proxy-client-react": "^4.1.2",
     "axios": "^1.6.7",
     "lodash": "^4.17.21",
+    "p-all": "^5.0.0",
```
Fewwy, line 44: Is this necessary? I don't see any imports of this module or any calls for the functions from this module.
I removed it while testing and I don't see any issues.

Fewwy (approved):
Other than one module I don't see any issues. Great job!

Fewwy (approved):

package.json
AsToNlele: Thanks, forgot about it

PR #722, opened 2024-03-27
fix(tests): using experimental strategy to detect flaky tests

Fewwy:
I found this https://docs.cypress.io/guides/references/experiments#Experimental-Flake-Detection-Features
In cypress docs and trying it


AsToNlele (approved):
let's try it, lgtm

PR #727, opened 2024-04-05
fix(RHINENG-8253): Update risk label feature flag

AsToNlele:
I wasn't aware this feature is behind a flag so adding it.
Includes a test that checks for labels and api requests if the feature is disabled.

How to test:
1. Run the app in `stage stable` with the flag `ocp-advisor.upgrade-risks.enable-in-stable` disabled in Unleash
2. Go to Clusters
3. Check that you don't see any labels with `Update risk` next to the Cluster name and that there's **no** api requests on `update-risks-prediction` endpoint
4. Enable the flag and check if you can see some clusters with the label, if not check the network tab if there's a request on the `update-risks-prediction` endpoint
5. If you don't have any `Update risk` labels, try the `prod-beta` environment, there might be some there

### DIFF
--- a/cypress/utils/interceptors.js
+++ b/cypress/utils/interceptors.js
@@ -184,6 +184,40 @@ export const featureFlagsInterceptors = {
       },
     }).as('getOcpWorkloadsFlag');
   },
+  upgradeRisksSuccessful: () => {
+    cy.intercept('/feature_flags*', {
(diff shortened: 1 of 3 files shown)


Fewwy (approved):
Good, very good, very very good! LGTM 
image

PR #730, opened 2024-04-08
fix: Update risk label second feature flag

AsToNlele:
There's a request for a second feature flag until the backend is sorted out.
The new feature flag is `ocp-advisor-ui-upgrade-risks`.
Should show the label only if it's enabled. To enable this feature both upgrade risks feature flags have to be enabled or in case of `preview/beta` env only this flag has to be enabled.


Fewwy (approved):
LGTM

PR #763, opened 2024-07-03
feat(RHINENG-8850): add objects name to the modal

Fewwy:
https://issues.redhat.com/browse/RHINENG-8850
You can test both versions of the table by changing objectsWithNames value from true to false.

New version of the table
image

### DIFF
--- a/src/Utilities/Workloads.js
+++ b/src/Utilities/Workloads.js
@@ -1,7 +1,7 @@
 import { SortByDirection } from '@patternfly/react-table';
-import { omitBy } from 'lodash/omitBy';
-import { cloneDeep } from 'lodash/cloneDeep';
-import { isEmpty } from 'lodash/isEmpty';
+import omitBy from 'lodash/omitBy';
+import cloneDeep from 'lodash/cloneDeep';
+import isEmpty from 'lodash/isEmpty';
 
 export const SEVERITY_OPTIONS = [
   {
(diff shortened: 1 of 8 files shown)


src/Utilities/Workloads.js
```diff
@@ -186,12 +188,28 @@ export const flatMapRows = (filteredRows, expandFirst) => {
 };
 
 export const passObjectsFilters = (objects, filters) => {
-  return Object.entries(filters).some(([filterKey, filterValue]) => {
+  console.log(filters);
```
gkarat, line 191:
```suggestion
```

src/Utilities/Workloads.js
```diff
@@ -186,12 +188,28 @@ export const flatMapRows = (filteredRows, expandFirst) => {
...
+  const cleanedUpFilters = _.omitBy(_.cloneDeep(filters), _.isEmpty);
```
gkarat, line 192: Please change all the lodash functions imports and don't use `_` since it loads the whole library.

src/Utilities/Workloads.js
```diff
@@ -186,12 +188,28 @@ export const flatMapRows = (filteredRows, expandFirst) => {
...
+        return (
+          filterValue &&
+          objects.uid.toLowerCase().includes(filterValue.toLowerCase())
+        );
+      /* case 'object_id':
         return objects.uid.toLowerCase().includes(filterValue.toLowerCase());
+      case 'display_name':
+        return objects.display_name
+          .toLowerCase()
+          .includes(filterValue.toLowerCase()); */
```
gkarat, line 210: I think we can safely remove that?

src/Utilities/Workloads.js
Fewwy: True, adjusted that

gkarat (approved):
Thanks @Fewwy

PR #764, opened 2024-07-03
feat(RHINENG-9189): adding back severity column to the workloads table 

Fewwy:
https://issues.redhat.com/browse/RHINENG-9189

Add severity column back to the main table
Turn on filter for severity column
Turn on sorting for severity column

To test this pr with more data variety for severity hits add mockdata into
src/Components/WorkloadsListTable/index.js
like this:

import React from 'react';
import { useGetWorkloadsQuery } from '../../Services/SmartProxy';
import { WorkloadsListTable } from './WorkloadsListTable';
import mockdata from '../../../cypress/fixtures/api/insights-results-aggregator/v2/workloads.json';

const WorkloadsListTableWrapper = () => {
const query = {
    isError: false,
    isFetching: false,
    isUninitialized: false,
    isSuccess: true,
    data: { workloads: mockdata },
    refetch: () => null,
  };
/* const query = useGetWorkloadsQuery(); */
  return <WorkloadsListTable query={query} />;
};

export default WorkloadsListTableWrapper;


src/AppConstants.js
```diff
@@ -431,9 +431,9 @@ export const WORKLOADS_LIST_COLUMNS_KEYS = [
 
 export const WORKLOADS_TABLE_CELL_NAME = 0;
 export const WORKLOADS_TABLE_CELL_RECOMMENDATIONS = 1;
-// export const WORKLOADS_TABLE_SEVERITY = 2;
-export const WORKLOADS_TABLE_CELL_OBJECTS = 2; // 3
-export const WORKLOADS_TABLE_CELL_LAST_SEEN = 3; // 4
+export const WORKLOADS_TABLE_SEVERITY = 2;
+export const WORKLOADS_TABLE_CELL_OBJECTS = 3; // 3
+export const WORKLOADS_TABLE_CELL_LAST_SEEN = 4; // 4
```
gkarat, line 436: nit-pick

```suggestion
export const WORKLOADS_TABLE_CELL_OBJECTS = 3;
export const WORKLOADS_TABLE_CELL_LAST_SEEN = 4;
```

gkarat (commented):
This works as expected, I can sort by the Severity column and filter out workloads by severities.

Though, I can see that the URL sort parameter indicates sorting by `objects` instead of `severity`

image

gkarat (approved):
Looks good 👍🏼

PR #817, opened 2024-10-21
fix(RHINENG-8333): correct ouia-type empty state

LightOfHeaven1994:
some tests are failing because FEC utils haven't been updated yet https://github.com/RedHatInsights/frontend-components/pull/2085


leSamo (approved):

LightOfHeaven1994:
> /retest

it will fail until ocp-advisor update frontend-components version as it contains some changes for cypress utils https://github.com/RedHatInsights/frontend-components/pull/2085

LightOfHeaven1994:
so finally I made it work, had to bump utils package too and added `redux-mock-store` package to dev dependencies otherwise cypress tests were failing locally and on Travis.
I also had to fix cypress tests for bulk select as structure looks correct now for bulk menu list and tests have to be updated

PR #898, opened 2025-02-14
chore(deps): fec and other package updates

Fewwy:


AsToNlele:
Cypress tests are failing

Fewwy:
> Cypress tests are failing

I cannot fix them for now, because stage is on PF6 and it broken OCP Advisor :( I need to fix it before getting back to this

PR #987, opened 2025-09-15
feat(RHINENG-19797): add lightspeed rebrand with feature flag

Fewwy:
https://issues.redhat.com/browse/RHINENG-19797
Added 2 messages for Lightspeed rebranding that are enabled when the feature flag is ON

How to test:
1) In the ClusterRules component, force the NoInsightsResults component to render with feature flag ON
2) In the ClusterListTable componen,t force the NoRecsForClusters component to render with feature flag ON

### DIFF
--- a/src/Messages.js
+++ b/src/Messages.js
@@ -663,4 +663,16 @@ export default defineMessages({
     id: 'updateRisksNotCurrentlyAvailable',
     defaultMessage: 'Update risks are not currently available.',
   },
+  noRecsFoundErrorDescLightspeed: {
+    id: 'noRecsFoundErrorDescLightspeed',
+    description: 'Recommendations table, cluster recommendations table',
+    defaultMessage:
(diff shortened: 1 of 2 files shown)


LightOfHeaven1994 (approved):
LGTM, thank you!

src/Messages.js
hugohezel, line 1: Found few places where old brand is still used :
- https://github.com/RedHatInsights/ocp-advisor-frontend/blob/master/src/Messages.js#L271
- https://github.com/RedHatInsights/ocp-advisor-frontend/blob/master/src/Messages.js#L518
- https://github.com/RedHatInsights/ocp-advisor-frontend/blob/master/src/Messages.js#L577
It might be that these messages are not used/obsolete though

hugohezel (commented):
Otherwise looks good to me, thank you @Fewwy!

